PR #1, opened 2019-11-20
add zh-CN locale

yushiro:


ClaudeMetz:
Thanks for this effort. Sadly, I can't merge this just yet. I'm not sure how to handle localisation of Factory Planner, as it has quite a few things to think about. I'll have to think about this some more. I really appreciate the effort though, and hope I can include it in the main mod at some point, but for now, you'll have to create a separate translation mod if you want to share this work.

PR #3, opened 2020-03-05
Create zh-CN/config.cfg

Yecgaa1:
Chinese config


ClaudeMetz:
Thanks a lot for your contribution. I decided to create a separate project for the localisation work, which can be found [here](https://github.com/ClaudeMetz/FactoryPlannerLocale). I included your contribution already, as a test case. If you want to make any changes to it, please create a pull request on that other repository from now on.
If you can, it would also be great if you could update the localisation file, as some things have changed in the meantime.

PR #10, opened 2020-12-02
Merging matrix solver

scottmsul:
I think the matrix solver is mostly ready to go. Please take a look at both the code and the dialog and let me know if you have any comments/questions. We could probably make some tweaks or better error handling, right now if there's a problem it just says "Linearly dependent recipes detected. Make sure the setup only allows for one solution."

I made it so the matrix dialog only opens when at least one free item is needed, which usually occurs with loops. For example one way to get this in base factorio is to create Light Oil from Coal Liquefaction, Steam from Solid Fuel, and Solid Fuel from Petroleum. Then you can select either Petroleum or Solid Fuel as a free variable for the byproduct, while the other gets eliminated.


modfiles/ui/dialogs/matrix_dialog.lua
```diff
@@ -26,20 +37,33 @@ local function create_item_category(modal_data, type)
...
+    local subfactory = ui_state.context.subfactory
+
     local split_string = split_string(element.name, "_")
     local type, proto_index = split_string[4], tonumber(split_string[5])
 
     local modal_data = data_util.get("modal_data", player)
-    local item_array = modal_data[type .. "_items"]
-    local swapped_proto = item_array[proto_index]
-    table.remove(item_array, proto_index)
+    local free_items = modal_data["free_items"]
```
ClaudeMetz, line 47: `free_items` isn't used in this function after it's assigned here. Is this a left-over line or does it point to a potential bug?

modfiles/ui/dialogs/matrix_dialog.lua
```diff
@@ -12,10 +23,10 @@ local function refresh_item_category(modal_data, type)
     end
 end
 
-local function create_item_category(modal_data, type)
+local function create_item_category(modal_data, type, label_arg)
     local flow_category = modal_data.modal_elements.content_frame.add{type="flow", direction="vertical"}
 
-    local label_title = flow_category.add{type="label", caption={"fp.matrix_" .. type .. "_items"}}
+    local label_title = flow_category.add{type="label", caption={"fp.matrix_" .. type .. "_items", label_arg}}
```
ClaudeMetz, line 29: What's the intention of `label_arg` here? The way it's setup currently, it doesn't do anything I think.

modfiles/ui/dialogs/matrix_dialog.lua
scottmsul: I think it was used in older version, it's removed now.

modfiles/ui/dialogs/matrix_dialog.lua
scottmsul: The `label_arg` gets used in `matrix_constrained_items` in `config.cfg`. Speaking of which, right now it's set to `Choose __1__ free items` - should we make it `free item` when `__1__` is 1 since it's singular?

ClaudeMetz: Oh my bad, I missed that that the locale string was changed, somehow. Disregard. Although yeah we can do that, I'll change it.

PR #11, opened 2020-12-12
matrix_solver: Determine prodded amount using crafts per second inste…

talshorer:
…ad of per timescale

In case a machine's speed would have it operate faster than one craft
per game-tick (1/60 second), the game would reduce this to one craft per
tick, with a different productivity value to error-correct. This
error-correction is taken into account by
`calculation.util.determine_prodded_amount`, which expects the amount of
crafts per second as its argument.

When we pass the total amount of crafts per timescale
(second/minute/hour), we can easily exceed 60 and kick the
error-correction code into effect. This would give wrong results for
minute/hour timescale, since the game is not going to error-correct in
the case of, say, 61 crafts per _minute_.

To address this, pass the amount of crafts per second, without
considering the timescale.

### DIFF
--- a/modfiles/data/calculation/matrix_solver.lua
+++ b/modfiles/data/calculation/matrix_solver.lua
@@ -563,25 +563,25 @@ function matrix_solver.get_line_aggregate(line_data, player_index, floor_id, mac
     if launch_sequence_time then
         time_per_craft = time_per_craft + launch_sequence_time
     end
-    local single_crafts_per_tick = timescale / time_per_craft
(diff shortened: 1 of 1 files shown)


scottmsul:
Thanks for finding this! I never really tested with high speeds or longer timescales so I never uncovered this. I guess my variable names were also incorrect/misleading.

With the fix and after some testing, I'm still seeing some disagreement between the two solvers if there's more than 60 crafts per second. I'll have to look into this more.

scottmsul (approved):

PR #15, opened 2020-12-13
Add done column allowing FP to be used with a todo list workflow.

Fiona:
Hi, I love FactoryPlanner a lot, it supplanted Helmod for me a long time ago. The new update is great, the updated UI is fantastic. 

During a recent Space Exploration play-through, I found myself really wanting to due FactoryPlanner like a todo-list. I want to check off aspects of a particular production chain as I complete them. I decided to add the functionality today on a whim.

As this is possibly a niche workflow it's implemented similar to the active toggle, so a default "false" preference. 
I realise that comments could accomodate this workflow, but that felt more like a workaround and wasn't nearly as nice to use for this.

Thanks :)

screenshot


ClaudeMetz:
Thank you for this, I'll have a look in the coming days.

ClaudeMetz:
Alright I had a look, and it looks good! I actually planned to add this functionality at some point in the future alongside other, much more involved new things, but there's not harm in adding it now I think.

You stuck to the existing structure and conventions of the mod really well. I am a super-stickler though, so I'll adjust some things. It'll look like a lot of changes, but be reassured that most of them are not because you did anything incorrectly, it's just that I apparently need to have things exactly my way.

So thanks, really appreciate the PR! Will get it into a release soon.

Fiona:
Thanks! I understand, I'm the same which I why I took so much care. Should I fix the conflicts or will you do that as part of your clean-up sweep? :)

ClaudeMetz:
No don't worry, I'll take care of that, there's a script that'll fix the migration issues I think, it's no big deal.

ClaudeMetz:
Alright, it seems I am totally unable to get this merged properly. My bad. Could you please address the migration issues yourself?

ClaudeMetz:
Whoops accidentally deleted the branch I changed this one to, go me.

ClaudeMetz:
Well I figured it out, no need for you to adjust anything now.

ClaudeMetz:
Alright I refactored some things, I'll change the wording and presentation a bit still before merging.

As a small note, you had some inconsistent formatting in there, like using tabs instead of spaces in some spots, and using a two wide indentation in the migration file. A properly configured editor would spot those for you, and make things easier in general, just as a suggestion :)

Regarding the implementation itself, I'm not 100% sure how you intended it to work. Should lines with subfloors have their own independent checkmark, or not? You implemented it both ways depending on the spot in the code we're talking about. It's not a big deal, I'm just curious what your intentions were. I totally understand that it's hard to see through the logic of the whole codebase, especially concerning the migration stuff, that stuff is complex.

Also, there's some potential for some 'smart stuff' here. Like automatically checking the checkmark of a line with subfloor if all recipes on that subfloor are checked, and so on. What do you think about those kinds of things? Would that help you, or would you prefer it to be just a 'dumb' list that the user fully controls himself?

Fiona:
I'm not a lua dev so I just ended up using some crappy editor. :P Sorry about that.

Hah I had thought about subfloors already and decided that it would be best if subfloor checkmarks were driven by the checks of the items in that subfloor. It seemed like that would be more useful.

ClaudeMetz:
No worries at all.

Can you expand on how exactly that would work in your opinion? Should the check on the parent of a subfloor mirror the first line of the subfloor, or maybe only be checked if everything on the subfloor is checked? If so, should it be clickable at all? If it is clickable, should it set all the checks on the subfloor to the one you 'click' it to be? Just curious how you'd like it to be since you're the one with the usecase for it.

Fiona:
I just had another think about it. My gut instinct was for the check of a subfloor line to only be checked if all the lines within the subfloor were checked off. On second thought, this functionality wouldn't be so clear.

In my revised opinion, a subfloor line should mirror the first line of the subfloor itself. Since that's essentially want a subfloor line is representing. The states of the checks on either side should probably affect each other when changed.

Although that assumes it's possible within the data structure for lines within a subfloor to recognise that they are infact part of a subfloor and determine which line of their parent they come from. Is that the case?

ClaudeMetz:
Alright that's the easiest-to-implement option for sure, I'll get right on that.

Yes that's intrinsic to the data pretty much. Lines with subfloors have next to no real data saves on them, they either get it from the first line of the subfloor or by the calculation model. It will work the same as the checkmark do disable a line then, where it's actually saved in the first line of the subfloor.

ClaudeMetz:
Just merged it, thank again for the PR! Appreciate it.

Fiona:
Yay! I'm glad to hear it. :) Thank you very much!

PR #16, opened 2020-12-14
GUI style crusade

raiguard:
I went through and adjusted the GUI styling / layout in a few places where I deemed a certain way to look better. These are all my personal opinion on things, so I don't expect everything to go through. Please let me know of any changes I should make or revert!

### Titlebar

image

The main dialog titlebar was standardized to fit other game windows, with corrected margins and paddings. Additionally, the drag target was changed from the drag handle itself to the entire titlebar flow, again to match vanilla GUIs. This makes it a bit easier to drag the GUI where you want it.

The modal dialogs can also now be dragged by the drag handles along the bottom.

### Pushed hovered/clicked styles

3ee10073-5211-4d65-bfce-21b3276e2490

For all buttons across the GUI that can be "pushed", unique hovered and clicked styles were added for them. This actually does not match the base game at the moment, because in 1.1 these styles were changed for some reason. Hopefully that change will be reverted soon. This makes it much easier to see that the button actually was pushed/unpushed when still hovering over it.

### Removed all rounded buttons

image

All of the rounded buttons were changed to be standard rectangular buttons. The only exception is the plus buttons, which retain the rounded shape.

### Dark choose-elem-buttons

image

All choose-elem-buttons in the mod were adjusted to use the `flib_standalone_slot_button_default` style, sized to 32. This matches the base game's choose-elem-buttons, making things just that bit more intuitive and familiar.

### Tabbed pane scroll pane style

image

The scroll panes in the tutorial dialog were updated to use the `flib_naked_scroll_pane_under_tabs` style, creating a more consistent visual separation when the scroll pane is active.

### Consistent subheader sizing

image

The production table subheader was significantly larger than the others across the game, so I resized it to match everything else. The size of the `Production` label was decreased to not look out of place, but it retains a slightly bolder font than other caption labels, letting it still stand out among the others.

### Major revisions to the picker dialog

image

Styling on the picker dialog was majorly revised to be more consistent with the base game item pickers. In fact, it's actually better-looking than the vanilla ones, which have some issues with scroll pane margins that I managed to solve for this one. :D

### Item box scroll panes

image

The item boxes had scroll panes, but they didn't actually work. I revised it to use the same scrollbar technique as in the picker dialog, and limited it to five rows.


ClaudeMetz:
Alright I finally got to playing around with this some. I'll tell you what I think about the various changes so you can amend/remove some of the commits like you said.

The titlebar changes are great, thanks for cleaning that up. Same with the pushbuttons, nice change.

I like my rounded buttons in some spots. Maybe one or two of them should be normal buttons, but I like most the way they are. The floor up/top ones and the mining prod override one should be rounded for sure. I could be talked into making the utilities and repair ones non-rounded.

I just can't agree with the dark choose-elem-buttons, I think vanilla is wrong on this one for my use-cases. It's an input field, and it should look like one (ie a textfield). It invites the user to click it if they want to input something. The contrast also is better. They are also easily confused with a disabled textfield in their dark form, which is not good.

The tabs change is great, I was pretty sure I was doing that wrong actually but didn't bother to fully figure it out.

The titlebar height was chosen intentionally to be larger, and I think it looks better. It's an important part of the interface, and it should be given some space to breathe imo.

The item picker looks good! Something is off about the height calculation though, because it shows a scrollbar when it [shouldn't yet](https://i.imgur.com/ADV3QWD.png), and is [off at the bottom](https://i.imgur.com/gwuhKU8.png) at the same time. Let me know whether you want to take a look at this or whether I should.

Finally, the item box changes are great too, thanks! Feel free to ask any questions of course.

raiguard:
I went ahead and reverted the dark choose-elem-buttons and getting rid of the rounded buttons.

I am pretty firm on my stance that the subheader should be consistent with the other ones around the game, but if you _really_ want it to be bigger, I can revert that as well.

I'll look into the scroll pane height thing for the item picker.

ClaudeMetz:
Alright do that, I'll stew on the toolbar height some more.

ClaudeMetz:
I fixed the picker issue. I also removed a line you labeled as a dirty hack, what was that for? Is it now also fixed by chance?

PR #19, opened 2020-12-16
Add wagon rate calculation

talshorer:


ClaudeMetz:
Thanks for this, I think I've come around to the idea somewhat. I'll try it out and see what I think.

ClaudeMetz:
Alright I had a look and refactored some things a bit. The only real change I made is that I removed the train carriage icon from the view state button, as I think it's a bit confusing to show the cargo wagon while both cargo and fluid ones are supported.

Otherwise, you did a really good job at integrating into the existing codebase. All these small changes are just because I'm a stickler, not because you did anything incorrectly. Thanks a lot for the PR, I'll get it merged soon, if you don't have any further objections or changes in mind.

modfiles/ui/dialogs/preferences_dialog.lua
```diff
@@ -200,7 +200,6 @@ local function handle_default_prototype_change(player, element, metadata)
 
     local modal_data = data_util.get("modal_data", player)
     if type == "belts" then modal_data.refresh.view_state = true end
-    if type == "wagons" then modal_data.refresh.view_state = true end
```
talshorer: Changing to a modded wagon with a different inventory size requires a refresh.

talshorer (commented):
I went over your refactor commits. I agree with the reorganization, the final result is neater.
My only real comment is the one about refreshing the view when changing the configured wagon.

PR #25, opened 2021-09-30
New solver

B-head:
For now, release it with only minimal UI changes.

### Features:

- Support production loops.
- Support for bottom-up design.
- Support machine limit.
- Respects the list in order and sub-floor structure like the traditional solver.
- No need to specify free items.
- (Calculations are slow.)

### Not yet implemented:

- Production percentage.
  - Plan to implement instead an limits will be supported for items.
- Product prioritize.
  - Plan to implement instead an external product and external ingredient as virtual recipe.
  - It is also improvement of the free items feature.
- From the reference item, add a recipe that consumes it.
- Predicting the amount of machines needed when switching machines.

### Changes:

- The solver is now specified using the drop-down menu.
- The default solver for new subfactories is now the interior point solver.
  - Solvers for existing subfactories will not be changed.
- Add name and timescale to subfactory_data.
- Hide the percentage column when using matrix solver.
- Add and change locale strings.
- In recipe_line, the item button is displayed even if the amount is zero.

### Additional note:

- The solver has been modified to specify a reference instead of a production goal.
- Reference is the minimum amount of items to use in a sub-factory.
It is used to specify the items you want to produce or consume.


raiguard:
I will try this in my SE playthrough and let you know if I find anything.

raiguard:
image

Well it almost immediately crashes when I load it in my SE save. I tried removing it then re-enabling it but that doesn't help.

It must be crashing in `on_configuration_changed` somewhere, because it doesn't do this on a new game.

[The Final Frontier.zip](https://github.com/ClaudeMetz/FactoryPlanner/files/7263778/The.Final.Frontier.zip)

B-head:
Existing saves need to be migrated.
We have changed the version of json.info to 1.1.26, please try again.

B-head:
Fixed.
@raiguard Please try again.

qdbp:
As far as I'm concerned, an LP solver is basically the raison d'etre of a factory planner for a pack like, e.g., Seablock. I would absolutely love to see this solver merged.

ClaudeMetz:
Yeah I agree it makes sense to have one, and I'm sorry I haven't really gotten to any of the proposals so far. I plan on working on the solver side of things at some point.

PR #26, opened 2021-10-02
Faster allowed free items search

RSBat:
Replaces checking every possible item for linear dependence with a single check on a transposed matrix. Might be unintuitive, but it works and I have a proof why (written version WIP).

Another performance boost should come from checking for zeros when reducing matrix. Might be even more important than the other improvement.

Also fixes any potential bugs caused by not rounding to zero on second elimination pass.


modfiles/data/calculation/matrix_solver.lua
```diff
@@ -236,6 +279,12 @@ function matrix_solver.get_linear_dependence_data(subfactory_data, matrix_metada
                 allowed_free_items[item_key] = true
             end
         end
+
+        print("Fast free items", fast_allowed_free_items)
```
scottmsul, line 283: Can you comment these out in the final version?

modfiles/data/calculation/matrix_solver.lua
```diff
@@ -203,13 +203,31 @@ function matrix_solver.get_matrix_solver_metadata(subfactory_data)
...
+    return transposed
+end
+
 function matrix_solver.get_linear_dependence_data(subfactory_data, matrix_metadata)
     local num_rows = matrix_metadata.num_rows
     local num_cols = matrix_metadata.num_cols
 
     local linearly_dependent_recipes = {}
     local linearly_dependent_items = {}
     local allowed_free_items = {}
```
scottmsul, line 229: For the purposes of this PR it's fine to delete the original `allowed_free_items` logic, save the fast result in `allowed_free_items`, and return the fast result.

modfiles/data/calculation/matrix_solver.lua
RSBat: Yes, everything below line 270 is old version + asserts for testing and should be removed

scottmsul:
I tested it out in my seablock game, everything worked ok. I was even able to make a subfactory that I couldn't before because previously it was too slow.

RSBat:
Here's an approximate proof of why this works:

Adding a linearly independent column is equal to adding a linearly
independent row to a transposed matrix. We want to add a row $x$ with
exactly one 1 at position $i$. Lets analyze the matrix after performing
reduction on all rows except our row $x$.

Forward:
If row is linearly independent => $i$ is in linearly dependent column.

Suppose it is not in a linearly dependent column. Then it is in a
column with a pivot element in row $y$. Because it is in a linearly
independent column, there are no other non-zero elements in row $y$.
So row $y$ matches our row $x$ exactly => our row is not
linearly independent.

Backward:
If $i$ is in a linearly dependent column => row is linearly independent

a) It is in a column without a pivot element => it becomes the
pivot element.
b) It is in a column with a pivot element in row $y$. Then row $y$ will
have at least one more non-zero element. So after subtracting row $y$
from our row $x$, the non-zero element in the row $x$ will be in a
different position. Repeat until it is in a column without a pivot
element. Note that each time the element moves to the right, so it
cannot go on forever.

PR #32, opened 2022-06-15
Code cleanup: Remove uncapped production ratio

curiosity-a:
The `uncapped_production_ratio` property is highly dependent on the inner workings of the sequential solver. Removing the property and making the affected features unreliant on it will make the code simpler and more solver-agnostic.

Side effect: the uncapped machine count being equal to the machine limit will count as hitting the limit. This change is unavoidable, as far as I can tell, but I do not consider the new behavior unreasonable.


ClaudeMetz:
Looking at this again, it is definitely a downgrade to user experience. I get that it is helpful/necessary for alternative solvers, but until one of those comes along, I'll keep this in. To be clear, this is not saying alternative solvers need to support it, I am open to dropping it for another solver, but until then I think it's better to keep it in.

PR #34, opened 2022-09-02
remove items within rounding error of zero in structures.class.add

scottmsul:
I looked into OneWheelDude's issue with -1's appearing as products. These only seemed to occur on subfloors, and upon further investigation I noticed the values were all really close to zero (typically around 10^-14 or 10^-15). I'm not really sure why they weren't being displayed as close to zero (maybe the UI can't handle scientific notation?). In any case, I found that `structures.class.add` should drop items that solved to zero, so I changed this code to check if the value is within 10^-10 of zero instead.


ClaudeMetz:
Turns out this was a UI bug, not a solver one.

PR #35, opened 2022-11-25
count only unfinished components in utility dialog

microlith57:
this PR adds `get_unfinished_component_data` functions to `Subfactory` and `Floor` that return only those components whose lines are not yet done.
it also modifies the utility dialog to use these functions (provided the checkbox column is turned on), so that the only components listed are those which are not yet placed.

(this can be implemented slightly more efficiently if #36 is merged)

### DIFF
--- a/modfiles/data/classes/Floor.lua
+++ b/modfiles/data/classes/Floor.lua
@@ -112,8 +112,9 @@ end
 
 
 -- Returns the machines and modules needed to actually build this floor
-function Floor.get_component_data(self, component_table)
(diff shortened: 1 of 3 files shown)


ClaudeMetz:
I am very sorry for not getting to this for almost a year. I finally have though. The code has become a bit outdated unfortunately, so I reimplemented it myself in the new system. Sorry that you won't get the credit for the change this way, but I appreciate it! Changed with [this](https://github.com/ClaudeMetz/FactoryPlanner/commit/0fa9d115c0ea19832b4103a22cc4d4f5097f0980) commit.

PR #37, opened 2022-12-07
Update to dictionary-lite

raiguard:
I think things should be good, but I wasn't able to test completely thoroughly.

dictionary-lite has a `handle_events()` function that adds all of the needed non-bootstrap handlers. I didn't use that here, instead opting to just replace the existing handlers and keep the `script` calls in factoryplanner.


ClaudeMetz:
Very much appreciated, thanks again!

PR #38, opened 2023-02-22
Simplify beacon UI when beacon overload is active.

ctgPi:
Space Exploration (and some other other mods that offer the same feature in isolation) rebalances beacons, by only allowing machines to be covered by at most one beacon at a time (machines under the influence of two or more beacons are disabled instead).

In this context, the "amount" field in the beacon selection UI is not necessary, as the only legal value for it is 1. This change cleans up the UI and forces the beacon amount to be always 1.

(This change keeps the "total" field, as that is still relevant for power consumption calculations.)

### DIFF
--- a/modfiles/changelog.txt
+++ b/modfiles/changelog.txt
@@ -6,6 +6,7 @@ Date: 00. 00. 0000
   Changes:
     - Added Ukranian translation (Thanks Met en Bouldry!)
     - Changed the ingredient combinator button to use a sprite instead of text
+    - Simplified the beacon UI when beacon overload is active
   Bugfixes:
     - Fixed that absurdly high module effects were not being capped correctly at the engine limit of +32767%
     - Fixed a few inconsistencies around raw ores and made them more visually distinct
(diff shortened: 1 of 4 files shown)


curiosity-a:
Should have just changed visibility and prepopulated the amount with 1, then no changes to logic would have been necessary.

ctgPi:
I didn't realize `ui_util` allowed for that — does setting `textfield_amount.visible = false` (and equivalently for the label) and `textfield_amount.text = "1"` do the trick? I can make that change on an additional commit.

curiosity-a:
Haven't tested it, personally, but I see no reason why it won't work.

And this has nothing to do with `ui_util`. This is just basic GUI API and common sense. The aim is to hide the beacon count. So just hide it and let the dialog logic be reused instead of unnecessarily complicating it.

ctgPi:
Implemented in [`0d1b352`](https://github.com/ClaudeMetz/FactoryPlanner/commit/0d1b352b5f65bee8a40f7051ba545db22e1ef5f5); I'll close the PR but keep the branch around (maybe better to copy the mod list, and the global-ish variable for the detection result?).

ClaudeMetz:
Thank you, will move over the mod list.

PR #149, opened 2023-10-30
Support fluid burners that produce steam.

falsedrow:
Specifically, py's Oil Burner MK 01.

### DIFF
--- a/modfiles/backend/handlers/generator.lua
+++ b/modfiles/backend/handlers/generator.lua
@@ -207,7 +207,8 @@ function generator.machines.generate()
             if fluidbox.production_type == "output" and fluidbox.filter
                     and fluidbox.filter.name == "steam" and proto.target_temperature ~= nil then
                 -- Exclude any boilers that use heat as their energy source
-                if proto.burner_prototype or proto.electric_energy_source_prototype then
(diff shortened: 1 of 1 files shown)


ClaudeMetz:
Err, don't think it works like that. This might add the entities, but it doesn't use them properly to calculate stuff, and more.

falsedrow:
Can you give me an example of what wouldn't work? There's already support for fluid burners in generate_category_entry.

ClaudeMetz:
Hm maybe I spoke too soon, I'll look into the details at some point then.

ljgw:
Funny, I ran into this issue as well, modified my local `generator.lua` in the same way (minus the line end..) and after seeing it work in Pyanodons, I came here to find that there was already this PR 😄

If it means anything, this code looks good to me, and works well in Py at least. 👍 from me!

Ifry:
@ClaudeMetz did the same thing as OP and @ljgw, and then found this PR
it works as expected for the py Oil Burner

ClaudeMetz:
Yeah I'll have to have another look at this.

ClaudeMetz:
I implemented fluid temperatures properly now, so I'll finally tackle this topic fully. Sorry that I never got to merging your version.

PR #190, opened 2023-11-24
Fix dataset iteration crash in migration 1.1.65

francoiscampbell:
Migration 1.1.65 would crash with this message because `subfactory` was the iteration index, not the value. (In my case, I was updating from 1.1.60 to 1.1.75)

Screenshot 2023-11-24 at 9 58 43 AM


ClaudeMetz:
Oof that's a silly bug, thanks for fixing it! Will release it soon.

PR #194, opened 2023-11-26
Include burnt results in lines, to support Py ash and canisters.

falsedrow:
A line's products are the recipe products plus the burnt result of any
fuel. This reveals the burnt result created by any line you already
have, so that you know to handle it. In the matrix solver you can
process the byproduct as usual (e.g. refilling fluid canisters).

What burnt result will be produced depends on the recipe, the machine,
and the fuel. This does not any recipes which fill demand for the
burnt result (e.g. you need ash, we don't suggest burning coal to make
steam in a boiler). This is pretty wonky, but it's an improvement from
not having ash show up at all. You can fake it by adding a one steam
as an output from your factory, and the extra steam will turn into
a byproduct. Or you can paste the ash-producing line you want in from
another factory and the solver will use it.

### DIFF
--- a/modfiles/backend/calculation/matrix_engine.lua
+++ b/modfiles/backend/calculation/matrix_engine.lua
@@ -628,7 +628,7 @@ function matrix_engine.get_line_aggregate(line_data, player_index, floor_id, mac
     local total_crafts_per_timescale = timescale * machine_count * in_game_crafts_per_second
     line_aggregate.production_ratio = total_crafts_per_timescale
     line_aggregate.uncapped_production_ratio = total_crafts_per_timescale
-    for _, product in pairs(recipe_proto.products) do
+    for _, product in pairs(line_data.line_products) do
         local prodded_amount = solver_util.determine_prodded_amount(product, unmodified_crafts_per_second, total_effects)
         local item_key = matrix_engine.get_item_key(product.type, product.name)
         if factory_metadata~= nil and (factory_metadata.byproducts[item_key] or free_variables["item_"..item_key]) then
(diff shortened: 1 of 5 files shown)


ThoreMehr:
When I use this branch Factorio locks up for ~8s when I open the gui.

falsedrow:
> When I use this branch Factorio locks up for ~8s when I open the gui.

Could you share a save file that shows this problem?

ThoreMehr:
https://mega.nz/file/8ZBHHbBA#WvjotqY7Jy9N-XFWsAdYfF345k7oGCJtrMP4BQoeEDc

It's a complete py run with a bunch of other mods.

falsedrow:
> When I use this branch Factorio locks up for ~8s when I open the gui.

@ThoreMehr The delay doesn't happen for me when I load your save. If it's still happening, you could try to get a stack trace. It's a bit of trouble to get it set up if you haven't already. Install VSCode and FMTK, run Factorio in debug mode, load the save (this takes a lot longer than outside of debug, that's normal, it took six minutes for me), hit the pause button in VSCode during the 8s delay.

ClaudeMetz:
Hello, sorry for only having a look at this now. I really appreciate the contribution. Unfortunately (and I don't blame you for this, the solver code is ass), your solution is quite complicated in approach, so I won't merge it. Nonetheless the next version will have this feature following my own implementation. Thanks again!

ClaudeMetz:
Oh you're totally right about the craft per second thing, big whoopsie. Will address that in a future version as well.

PR #321, opened 2024-08-11
Prevent shifting Lines above the top element in a subfloor

### LINKED ISSUE #312: "send to top" on subfloor replaces that subfloor's final product
### Problem Description

Usually the subfloor is restricted to produce its original product, that product is locked at the top and can't be moved down. 
However, using ctrl+click to move another product to the very top will replace it and change what the subfloor is trying to produce. Only way to fix it is to move the original product down farther and then send it to the top.

### Reproduction

_No response_

mattnotmitt:
Fixes #312.

This was a bit of a slapdash solution which doesn't preserve the nice encapsulation that the `Object` -> `Floor` inheritance provides. If this is acceptable, then great! 

However, I was considering the possibility of adding a `.locked` field to `Object` which could be applied to the top level `Line` in a subfloor, which could be checked by `_shift` without having to know about the structure of floors and lines.

Let me know what you think!

### DIFF
--- a/modfiles/backend/data/Line.lua
+++ b/modfiles/backend/data/Line.lua
@@ -9,6 +9,8 @@ local SimpleItems = require("backend.data.SimpleItems")
 ---@class Line: Object, ObjectMethods
 ---@field class "Line"
 ---@field parent Floor
+---@field next LineObject?
+---@field previous LineObject?
 ---@field recipe_proto FPRecipePrototype | FPPackedPrototype
 ---@field production_type ProductionType
 ---@field done boolean
--- a/modfiles/changelog.txt
+++ b/modfiles/changelog.txt
@@ -3,10 +3,13 @@ Version: 0.00.00
 Date: 00. 00. 0000
   Features:
     - Added support for burnt items being output as burner fuel is used up (Thanks falsedrow!)
+    -
   Changes:
     -
   Bugfixes:
     - Fixed that beacon module defaults would reset themselves when the active mods changed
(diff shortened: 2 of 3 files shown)


ClaudeMetz:
Actually, I'm sorry, I didn't look at this properly enough, this is not actually the spot I would put this. Since this is specific to lines/subfloors, which already has some specific handling around the 1st subfloor line-awkwardness, I'd put it there and not at the object layer.

I'd handle it before calling the Object shift method, limiting the spots_to_shift number manually to avoid the issue. Handling is done in [this](https://github.com/ClaudeMetz/FactoryPlanner/blob/master/modfiles/ui/main/production_handler.lua#L6) method.

Do you want to have a go at that, or should I? :)

mattnotmitt:
@ClaudeMetz happy to have a go. Will hopefully get around to it before the weekend.

ClaudeMetz:
Alright great, happy to help with anything if need be.

mattnotmitt:
Given it a shot. Not sure it's the most idiomatic Lua but it seems to work! I also did some type-hinting as FMTK didn't seem all too happy and added some blank lines to the changelog to avoid merge conflicts in the future.

ClaudeMetz:
Looking good, thanks for the patience. I basically ignore the changelog warnings as my release script cleans things up anyways. And like you might have noticed, I haven't transitioned to using annotations everywhere quite yet, so it's a bit of a mess of warnings.

Anyways thanks, will do a with this included release now.

Issue #88, opened 2023-07-25
Solver bug with catalysts


ClaudeMetz:
I haven't investigated this bug too much yet, but it's related to catalysts. They are probably handled a bit weirdly by the mod, so I'll see if I can come up with a clearer system for them.

<img width="821" alt="image" src="https://github.com/ClaudeMetz/FactoryPlanner/assets/22834555/4ac1c371-bde8-45cc-9c91-a2f692972db2">

heinwessels:
Relavent issues:
https://github.com/ClaudeMetz/FactoryPlanner/issues/255
https://github.com/ClaudeMetz/FactoryPlanner/issues/257

xane256:
I just ran into this issue again with vitamelange nuggets, and I think the issue is in this line https://github.com/ClaudeMetz/FactoryPlanner/blob/15be116ec27aeec0b1d2c3c1c3c0f2c41bdda6db/modfiles/backend/handlers/generator_util.lua#L30 which causes `prodded_amount` to become negative when it subtracts the catalyst amount of stone (4) from the average value of the range (2). The analysis shows an 80% productivity debuff to the stone output instead of an 80% buff when using 4 P9 modules (+20% each).

![image](https://github.com/ClaudeMetz/FactoryPlanner/assets/71676379/83f3d935-f79c-48f7-88cc-0c505411c33c)
![image](https://github.com/ClaudeMetz/FactoryPlanner/assets/71676379/d00fefbd-3df1-466b-b0d7-7dc9b79223de)
![image](https://github.com/ClaudeMetz/FactoryPlanner/assets/71676379/9a4e8377-7402-417a-bb64-e8f71de76523)
![image](https://github.com/ClaudeMetz/FactoryPlanner/assets/71676379/90262674-f186-47a5-acb5-5ee2c3d7a846)

Barring a solver rewrite, a possible quick fix may be to clamp `(base_amount - (base_item.catalyst_amount or 0))` to be non-negative.

Thank you for your time!

Scarabytes:
It seems like this issue appears when a catalyst item with a range of possible numbers is used as an output but doesn't appear as an input.
Iirc, the "product" type item entries for the recipes already have the corresponding inputs subtracted from them before, so it only can become negative in this very case (or if a mod author set the catalyst item too high).
In the SE case outlined above, xane256's solution would work.
[...]

PR #323, opened 2024-08-20
fix prod calculation for random output amounts with catalyst specified

kalenedrael:
Productivity only applies to amounts above the catalyst amount. When the productivity bar fills, a set of outputs (random or not) is generated and the catalyst amount is subtracted from each output, with a minimum of 0 (no negative outputs). This fixes the calculation of the "proddable amount" - the effective amount of output that prod can apply to.

This behavior was tested using [this mod](https://mods.factorio.com/mod/prod-recipe-test). Compare expected and actual results in [this spreadsheet](https://docs.google.com/spreadsheets/d/1Xj5CfOe58bS2IqXXUqWxfcjYPRmfYRuCLOoIiOSxjSQ/edit?usp=sharing).

Fixes issue #88.

### DIFF
--- a/modfiles/backend/handlers/generator_util.lua
+++ b/modfiles/backend/handlers/generator_util.lua
@@ -19,15 +19,39 @@ local generator_util = {}
 ---@return FormattedRecipeItem
 local function generate_formatted_item(base_item, type)
     local base_amount = 0
+    local catalyst_amount = base_item.catalyst_amount or 0
+    local probability = (base_item.probability or 1)
+    local prod_part = 0
+
     if base_item.amount_max ~= nil and base_item.amount_min ~= nil then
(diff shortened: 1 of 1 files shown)


ClaudeMetz:
Hello, appreciate the very nicely done pull request! I won't be able to merge it for a 1.1 release as this code has changed a lot for the internal 2.0 branch and merging it will be a mess. 1.1 releases are basically done as 2.0 is so close.

Would you be okay with revisiting this once 2.0 is out? It probably won't be too difficult to adapt it to the new stuff.

kalenedrael:
I can do that. Any chance I can get access to the 2.0 branch? I can wait otherwise.

ClaudeMetz:
Unfortunately not, you can't make a private fork of a public repo for some stupid reason and setting up a separate one is a PITA.

PR #363, opened 2024-10-24
Fix issue that the recipe productivity bonus does not update when set…

Dutch0903:
… to 0


Dutch0903:
Solves issue https://github.com/ClaudeMetz/FactoryPlanner/issues/364

ClaudeMetz:
Thanks a lot!

Issue #417, opened 2024-11-21
Module plaste in same place crash (maybe just in 2.0.12)


karl0-dev:
### Problem Description

The mod Factory Planner (2.0.12) caused a non-recoverable error.
Please report this error to the mod author.

Error while running event factoryplanner::on_gui_click (ID 1)
__factoryplanner__/backend/data/Module.lua:125: attempt to index field 'parent' (a nil value)
stack traceback:
	__factoryplanner__/backend/data/Module.lua:125: in function 'validate'
	__factoryplanner__/util/clipboard.lua:48: in function 'paste'
	__factoryplanner__/ui/main/production_handler.lua:124: in function 'handler'
	__factoryplanner__/ui/event_handler.lua:173: in function <__factoryplanner__/ui/event_handler.lua:121>

### Reproduction

factory planner 2.0.12

1. Open factory planner.
2. Do factory planning
3. shift right click on modules -> shift left click on same spot
4. crash

edit1: actually happens when you paste in other places too.
edit2: reverted to 2.0.10 it doesn't crash anymore

video: Expires in 2 days from post
https://steamusercontent-a.akamaihd.net/ugc/43442168960839811/8799CB6938EF1EEC6F7B9D96AE1D70A466F4E5E5/

save file: (onedrive)
https://1drv.ms/f/s!AucteQ6GRbR-2RbGgiv0tqCNue_C?e=A0geMz

durandal42:
Also happens when pasting modules into a different module slot, sometimes:

![Screenshot 2024-11-21 193901](https://github.com/user-attachments/assets/5eb3990d-d0a9-4673-9791-e6eb258413d0)

jmklix:
I also got this error and fixed it with this PR: https://github.com/ClaudeMetz/FactoryPlanner/pull/419

ClaudeMetz:
Thanks for the report, this is fixed for the next release!

PR #419, opened 2024-11-22
fix crash when values are nil

jmklix:
Fix [this](https://github.com/ClaudeMetz/FactoryPlanner/issues/417) crash by checking if self.parent exists so it doesn't crash when that is `nil`


ClaudeMetz:
Thanks for the quick PR! Unfortunately, it's not quite the right solution. It addresses the symptom, but doesn't treat the cause, if you will. This is the fix I did instead: a72cf8cff23f597981aa2c583aea7ba89c4efd41

PR #457, opened 2024-12-24
[WIP] Proof-of-concept test harness using a separate mod

xaviershay:
Still haven't got my head around best way to actually control the planner - see in-progress-but-wrong code in the `reset` function. Think this is mostly just me not understanding the data/control structures.


ClaudeMetz:
Hello, so sorry for not getting back to you earlier. I'll try to have a look soon.

PR #476, opened 2025-01-18
Support recipe products with no name

danielmartin0:
Factory Planner crashes in 2.0 when paired with mods that use the new feature https://lua-api.factorio.com/latest/types/ResearchProgressProductPrototype.html. This product causes the recipe to contribute towards a research in progress.

This PR makes Factory Planner ignore recipe products that have no name, such that its existing functionality works as intended without crashing.


ClaudeMetz:
Okay I did a different fix to yours (https://github.com/ClaudeMetz/FactoryPlanner/commit/6da34d47253b311a065737396f54abf8748c7527), which only touches one place in the code. It means the recipe can't be put into FP (not sure if it could with your solution), but yeah it's a simpler fix. I'll likely do proper support for these kinds of recipes at some point to properly adress it.

PR #518, opened 2025-04-16
Super ultrawide support

NIronwolf:
Adds some additional narrower width percentage selections to get a narrower compact view on super ultra-wide monitors (32:9) that is comparable to the widths that the existing list makes on standard 16:9 monitors.

Note: I'm not positive increasing the minimal_width in the style is the best solution, but without the change the scroll bar makes all the added choices get an ellipsis and not show the whole number.

### DIFF
--- a/modfiles/control.lua
+++ b/modfiles/control.lua
@@ -43,7 +43,7 @@ GLOBAL_HANDLERS = {}  ---@type { [string]: function }
 
 PRODUCTS_PER_ROW_OPTIONS = {5, 6, 7, 8, 9, 10}
 FACTORY_LIST_ROWS_OPTIONS = {20, 22, 24, 26, 28, 30, 32}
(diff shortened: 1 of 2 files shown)


ClaudeMetz:
Ah this probably makes sense. I think it should only be jumps of 2 percent though for the new numbers, just like the old ones. That'll be precise enough.

I'll have a look at the styling issue, it's probably the way to go for that weird interaction.

NIronwolf:
I went to 1 steps, because being double wide, it makes the same amount of "pixels" per step. :)

ClaudeMetz:
Yeah but still, it's just too many I feel.

NIronwolf:
Should I work up a more complicated patch with a setting that let's you pick your monitor type and presents a halved valve list and is back to not needing to scroll?

ClaudeMetz:
Nah no need to overcomplicate it I think, just do steps of 2 for everyone and it'll be fiiine.

PR #523, opened 2025-05-02
Fix amount requested via logistic network

marfenij:
Hi, in case when you have some of items in inventory - request only missing items not work as expected:

e.g. - you need 4 Assemblers and you already have 2, prior behavior request 2 and you already have them, so you got nothing

so requesting whole count fix that case


marfenij:
Hi @ClaudeMetz,
any objections ?

ClaudeMetz:
Hello, I haven't had the time to look at this yet, I'll get back to you when I have done so.

PR #540, opened 2025-06-22
Super ultrawide

NIronwolf:
My apologies, this is my first time actually submitting a pull request.  I mistakenly thought that you could/would remove the extra width choices.

I've pulled the odd widths.

I appreciate letting me learn a bit about collaborating.


ClaudeMetz:
Hi, sorry about closing the other one then, I wasn't aware you were new to this whole thing!

In general, I'd avoid modifying the PR myself, since it's your idea and contribution, so you should get the credit for it. It's not me being unable/too lazy to change it, it's just that it's your work so I don't want to interfere too much, if that makes sense.

However, there's still some uneven numbers in there now 😅

Issue #348, opened 2024-10-20
Matrix solver doesn't support burnt_result


ClaudeMetz:
No idea how to make it work with the solver, I tried but it caused other issues.

Also, fuels from subfloors aren't shown on its parent line, but are part of the total. Probably just not set properly on the line.

HildarTheDorf:
I assume this is half the bug I just found as well.

With the traditional solver, Full Py automation science plan was requesting 5.8 ashes, 1.5 raw coal (only used as fuel), and producing 20.1 ashes as byproduct.
With the matrix solver it was requesting 6 ashes, and reporting no byproducts.
Seems like it also wasn't requesting fuel as well as ignoring the burnt_result byproducts.

ClaudeMetz:
It should have been requesting fuel, but yeah it won't show ashes at the moment. Also, if you have coal-producing recipes in your plan, it might just cover the fuel-coal internally, thus no external requests.

ClaudeMetz:
Solved by #546

PR #546, opened 2025-06-29
Matrix solver updates 06/2025

scottmsul:
- adds burnt_result to a line's byproducts if present
- fixes fuel not displaying properly as an ingredient for higher floors / factory overview


ClaudeMetz:
Thanks a ton, this'll make people very happy :) Solves #348

PR #579, opened 2025-07-18
fixed bug where burnt_result was always set to byproduct in the matrix solver

scottmsul:
The `burnt_result` feature was not implemented correctly in the matrix solver, as it was always being set to a byproduct. Whether the burnt_result is a product or byproduct depends on burnt_result's context within the matrix solver. It should always be set to product unless it is an explicit byproduct or free item. This matches the behavior for how normal recipe outputs also classified.


ClaudeMetz:
Thanks a ton as always Scott!

Issue #222, opened 2023-12-20
Easy to get locked up when selecting unrestricted items


alethiophile:
### Suggestion

The current interface for selecting unrestricted items has some land mines in it. Often when dealing with complex recipes it's not immediately clear which items should be unrestricted, and it requires some trial and error to see which ones make the result come out reasonable.

However, there are a few issues here. One is that the matrix solver may occasionally find a solution in which one or more of the provided recipes runs backward (negative coefficient). This is obviously just an error, and generally means you've suggested an unrestricted item that doesn't make sense; it would be good to detect this and disallow selections that lead to a negative number.

More seriously, if you get the "linearly dependent recipes" error, this removes your ability to select unrestricted items. Thus, the only way to fix it once you get that error is to remove one of the relevant recipes, add it back, and then make another selection. It also clears your existing selection, meaning that if you've got a working recipe and you get this error while modifying it, it can take some work and guessing even to get back to your working recipe.

You should be able to select unrestricted items like normal even after a "linearly dependent recipes" error, and making a new selection should rerun the solver again. This would allow trial and error in selecting items without fear of messing up your existing arrangement.

ClaudeMetz:
Thanks, appreciate the well thought-out suggestion! Not showing things that lead to negative numbers is tricky, since it would require computing all the permutations to find that out, to my understanding. That is not really feasible, performance-wise unfortunately.

The other suggestions are great, I'll look into them when I get the chance :)

jgoldshlag:
+1 for this one for me. Playing Nullius, I often choose the wrong unrestricted item, and having to delete a line and then re-add it to re-select the unrestricted items is a pain. I think the old UI used to let you re-choose the items, a simple button to clear unrestricted items and force me to re-select them again would be great

ClaudeMetz:
This was long ago, but version ` 1.1.73` changed this UI. Does it still pose problems for you @jgoldshlag or is it okay now?

jgoldshlag:
It is still an issue, there doesn't seem to be a way to back out of a selected unrestricted item when you choose wrongly. Still running into this in my Pyanadons run

ClaudeMetz:
Okay thanks for letting me know, I'll look into it.

puggan:
Still a problem in v2.0.9.

My work around is to switch solver to Traditional and back to Matrix.
It resets the "unrestricted" and get you out of "linearly dependent recipes".

[...]

Issue #85, opened 2023-07-25
Improve matrix solver UI


ClaudeMetz:
The matrix solver UI is pretty unintuitive, and there's ideas to improve it that I should be able to do. Below are some suggestions by users.

> So I am now aware the matrix solver is probably not adjustable, but depending on the interface with it, some of these might be feasible.
UI - In the popup, give details on the ranges for each possible resource, if it will increase or decrease, and specifically let you choose one of those. I don't mind producing extra petroleum, but I can't exactly consume more of it since I have zero.
Linearly Dependent - When this goes off from you selecting the "Wrong" unrestricted items, you can't unselect that last unrestricted item, but must either remove recipes or or turn the matrix solver off, which removes any void recipes or previously chosen unrestricted items.
Also, I would prefer being able to have a list of those recipes and be able to say "I prefer Recipe A over Recipe B, use A as much as you can then switch to B" or something like that.
Isolation. Subfactories, currently, are mere organization tools. They only make things look nicer, by hiding complexity, but don't actually act as seperate subfactories. Suggestion that the inside of a subfactory is treated entirely like a encapsulating class, such that its insides are completely blackboxed to any higher level factory. So for instance, you put a Koverox process in there, matrix solver on, and it figures out the loop. You go back up a level, and it just treats that subfactory as a "virtual recipe" which consumes 2 depleted uranium to make 1 enriched uranium.
Locking. Allow recipes to be "Locked" such that the matrix solver can't touch them, but has to work around them. As far as it is concerned, a locked "Gear" recipe just adds a demand to the system of producing 2 iron and consuming 1 gear.
Minimax. - Instead of absolute values of whatever, have ranges. All resources are +- infinite by default, but if you demand an output, it defaults to a minimum value given, You can lock that by adding a max. Inputs are the same, where you are inputing at least this amount. The matrix solver then runs inside those constraints, which you can narrow down until it is trapped at one chosen solution.

ClaudeMetz:
I improved the ergonomics a bit with [this commit](https://github.com/ClaudeMetz/FactoryPlanner/commit/4f839128148e2d74299572366c347aee55f359d5), but there is more that could be done.

ClaudeMetz:
[...]

PR #586, opened 2025-07-23
updated matrix solver to preserve selected unrestricted items when the factory enters a linearly dependent state

scottmsul:
Oftentimes when using the matrix solver, the factory can enter a linearly dependent state. This happens when certain columns are mathematically redundant.

Columns in the matrix solver can represent recipes that solve to a building count for a given line, or items that solve to a factory's net input or output. When a factory becomes linearly dependent, typically only a small subset of columns are part of this linear dependent group. The UI can be designed to show these problematic recipes or items to the user so they can be informed of how to resolve it.

A common workflow when using the matrix solver is to eliminate the number of net ingredients or byproducts to help balance the factory. Examples might include oil cracking to remove heavy or light oil byproducts, kovarex to eliminate both U-235 and U-238, or voiding in certain mods.

If a recipe is added to eliminate a previously selected unrestricted item, *technically* this always results in linear dependence. To resolve this, the current logic is to always wipe the unrestricted items and request the user to re-select them, as this preserves the recipe (which the user usually wants) and removes the problematic unrestricted item. However, this is overkill since it forgets all of the user's other choices from previously selected unrestricted items. For vanilla or small factories this isn't too big of a deal, but has received complaints from users playing larger mods with complex setups such as Pyanodons. 

The new logic requests the user to resolve issues in the following order:
1) Unselect any linearly dependent unrestricted items
2) Remove any linearly dependent recipes
3) Select any required unrestricted items

For the first option, when a recipe results in only one possible item, the code will automatically select it. For simple setups like cracking or kovarex mentioned above, the workflow should be identical to before. When two or more choices are possible, it uses the bottom of the production box for this selection.

One way to test the new UI functionality is with the following factory:
- Output plastic, solid fuel, and lubricant (I used 100 for each)
- Add three recipes to produce these, in order to cause petroleum, light oil, and heavy oil to become ingredients
- Add advanced oil processing with light oil and heavy oil as unrestricted items (petroleum will then be eliminated)
- Add heavy oil cracking -> this will force a decision to remove either heavy oil or light oil from the unrestricted items!

Note I only added the english versions for these captions which would need translations.


scottmsul:
**Issues**

- [Issue 85](https://github.com/ClaudeMetz/FactoryPlanner/issues/85) contains many ideas, one of which is this update, so only partially resolved. 
- [Issue 222](https://github.com/ClaudeMetz/FactoryPlanner/issues/222) is really two issues in one. The first is requesting to not display unrestricted items that result in negative solves. The second is this issue. This one would be half-resolved.
- [Issue 645](https://github.com/ClaudeMetz/FactoryPlanner/issues/645) requested this exact feature and would be fully resolved.

PR #593, opened 2025-08-02
Suggestions from last recipe

Master-Killer:
Prominently display the ingredients and products from the most recently added recipe in a separate list "From last Recipe", so you don't have to hunt for them.


ClaudeMetz:
I'm sorry for never getting back to this. I don't think this is something I want at this point, sorry.

PR #624, opened 2025-09-25
Break matrix solver toggle into its own boolean

ClaudeMetz:


ClaudeMetz:
Alright I'll double-check things, appreciate the quick response.
The reason I think it's never called is that `factory_data.matrix_free_items == nil` means the matrix solver is disabled, so it wouldn't run this code to solve things. It could run in one of the pre-solve checks but from what I saw it doesn't get to there with it being `nil`.

PR #740, opened 2026-04-17
updates to help matrix solver with numerical precision

scottmsul:
This fixes a user-reported linear dependence bug from uranium refining in pyanodons. The issue was a side effect from adding energy as a solver item. Basically this particular recipe requires >1GW per building, which created a matrix whose minimum entries exceeded the tolerance of 10^-10 x (max value), in this case around 0.1. This PR includes the following updates:
1. Set tolerance to 1e-12 from 1e-10 (Lua uses doubles so I think this should be ok)
2. Scale free variables such that 1 unit equals its max value per craft
3. Rescale rows such that the max crafting value is 1

### DIFF
--- a/modfiles/backend/calculation/matrix_engine.lua
+++ b/modfiles/backend/calculation/matrix_engine.lua
@@ -316,11 +316,12 @@ function matrix_engine.get_matrix_data(factory_data)
     local matrix = matrix_engine.get_matrix(factory_data, rows, columns)
 
     return {
-        matrix = matrix,
+        matrix = matrix.matrix,
         rows = rows,
         columns = columns,
         free_variables = free_variables,
         matrix_free_items = matrix_free_items,
+        free_variable_scale_factors = matrix.free_variable_scale_factors
     }
 end
 
@@ -332,6 +333,7 @@ function matrix_engine.run_matrix_solver(factory_data, check_linear_dependence)
     local columns = matrix_data.columns
     local free_variables = matrix_data.free_variables
     local matrix_free_items = matrix_data.matrix_free_items
+    local free_variable_scale_factors = matrix_data.free_variable_scale_factors
 
     matrix_engine.to_reduced_row_echelon_form(matrix)
     if check_linear_dependence then
@@ -355,6 +357,11 @@ function matrix_engine.run_matrix_solver(factory_data, check_linear_dependence)
             end
         end
         return linearly_dependent_variables
+    end
+
+    -- rescale ouput column based on free variable scale factors
+    for idx, scale_factor in pairs(free_variable_scale_factors) do
+        matrix[idx][#columns.values+1] = matrix[idx][#columns.values+1] * scale_factor
     end
 
     local function set_line_results(prefix, floor)
@@ -570,6 +577,7 @@ function matrix_engine.get_matrix(factory_data, rows, columns)
         local col_str = columns.values[col_num]
         local col_split_str = util.split_string(col_str, "_")
         local col_type = col_split_str[1]
+        -- note this string "item" is an internal matrix-solver convention and is unrelated to item types
         if col_type == "item" then
             local item_id = col_split_str[2].."_"..col_split_str[3]
             local row_num = rows.map[item_id]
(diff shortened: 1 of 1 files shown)


ClaudeMetz:
Thanks as always! Seems to work great. I ran a super basic performance test and it does seem to be 10-20% slower, but not sure there's anything to do about that. The matrix solver is already 5x slower than the traditional one, and I might need to find a way to mitigate that anyways.

scottmsul:
I hadn't even thought about performance, and have given essentially zero thought to optimization. If you want this is something we could look into, there might be a lot of low-hanging fruit here. How big does a factory need to get before running into noticeable slowdowns? Have any users reported any issues?

ClaudeMetz:
Eh it's kind of hard to quantify performance, as it's very situation dependent. The matrix solver currently takes around 10ms on my machine for a small-medium sized plan (ignoring GUI refresh here; the normal solver is at around 2ms for comparison). A frame is 16ms, so theoretically to not have any dropped frames, the rest of the game would need be faster than 6ms. That's the case in a lot of saves, but a lot of saves are also slower than that, which would cause stutters in theory. In practice though, these GUI things don't stutter as much/quickly as you'd expect, I'm not sure why.

Additionally, the solver runs a lot of course. It runs on every change to a factory, and during migrations, sometimes when settings change, etc. It means the solver sometimes runs for all factories at once, which stack up the performance hit. Those latter ones can be mitigated by spreading things out, but I haven't done that yet.

Speaking of, one potential avenue for performance improvement would be to spread the solving over multiple ticks, which means less processing per frame and thus less stutters. I assume the matrix solver process can be started and stopped, but it's probably more effort than it's worth. The important part would be to know the remaining calculation effort at certain points, so you can split it up depending on effort required. Not sure if that's a thing with this kind of solver. Alternatively it could be split based on matrix size, if that correlates to solving time?

As for 'normal' performance improvements, I'm sure there's quite a bit that could be done. I haven't really optimized my solver either, except by picking sensible complexity algorithms as far as I remember. Although it was adapted and extended so much over time, there are probably inefficiencies nowadays.

All that said, there aren't really any complaints about performance. I'm not quite sure why, like I said I think the game hides it better than you'd think from the raw frame times. Esp. in multiplayer it should be noticeable if you're stuttering while playing because someone else is using FP. But I don't play multiplayer, and I have a fast system otherwise so performance is always great. Maybe people are just okay with the slight hitches, and expect mods (especially complex ones) to not be super smooth. Still it's a bit of a crappy user experience.

Lastly, I don't know if you know (don't remember if we talked about this), but there's a debugger extension for VSC for Factorio. It can also do profiling on a per-line basis, so it's something that could be helpful to play around with if you want. Otherwise there's the simpler build-in `game.create_profiler()` function.

durandal42:
> All that said, there aren't really any complaints about performance. I'm not quite sure why

I have made factory plans large enough that performance became noticeable, e.g. "in SpaceExploration, all Norbit space sciences in a single factory" which had hundreds of lines (maybe a thousand?).

It never occurred to me to complain about performance, because I realized I was doing something heavy and stupid, and I was very pleased simply to be allowed to keep doing it. Frame stutters when updating a factory plan did not meaningfully impact my gameplay (which at the time consisted of staring at a literal spreadsheet, so it's not like I needed smooth framerates).

All this is to say, while I won't argue *against* performance improvements in the solver, they're not high on the list of improvements I want.

durandal42:
The worst of it was when the whole factory was covered in beacons with a mix of speed/efficiency (~nothing made in space supports productivity, so it was purely optimizing for small footprint), and I'd get a module tier upgrade. The workflow then went:
1) copy the factory, give the new one a name suffix that says what the new tech being applied is - in this case, a new tier of speed module
2) in the new factory, update one beacon
3) copy-paste that beacon down the line across the entire factory

I don't know if step 1 triggered a matrix solve, but 2 did (as expected) and 3 did on every click (also as expected, but perhaps not as desired). I felt it, but the sheer pace of Factorio:SpaceExploration is such that losing a few milliseconds to planning an upgrade that was going to take me hours to deploy just... didn't bother me that much.

In this worst case scenario, given the choice between "20% faster matrix solve" or "run the solver once when I've finished updating the factory", I would probably not notice the former, but the latter would have felt like a big improvement.

So, some ideas I'm chewing on:
- run at most one solve per tick (so in cases where all factories need an update at once, you don't spike all the work in a single tick)
- any single factory which takes longer than N ms to solve gets flagged as "big"
- "big" factories don't update until the player clicks a button that says "solution is out of date; click here to recalculate (might be slow)".  (can you tell I'm not a UI developer? 😭 )

ClaudeMetz:
Appreciate all the insight. It's reasonable to expect performance to suffer with large plans for sure, my concern is mostly about multiplayer where you'd hitch other players while they are playing the game itself, which could get very annoying. I think latency hiding helps with it, but there's obviously limits to what becomes just too slow.

As for your ideas, they would be nice to have. The issue is that the rest of the code expects that any factory is always solved up to date. Doing any delayed solving that still allows viewing and editing of a factory is going to trip up the existing code. That is theoretically solvable, but it'd probably be very annoying to suss out, and clutter up the code in many places. I guess it would be an opportunity to write nice tests for all this as the code is updated, but yeah I'm not sure if it's worth it at all unfortunately.

Issue #332, opened 2024-10-13
Add interface for content mods to improve compatibility


ClaudeMetz:

DeadMG:
An excellent improvement for SE compatibility would be to ignore Arcospheres in all factory plans, and perhaps add support for core miner yield calculators

ClaudeMetz:
I just added the first compatibility interface to the mod (not yet released) for #752. So I'm happy to look into more now. @DeadMG if you have any specific ones, I'd appreciate you opening a new issue and explaining the details of what would be required!

ClaudeMetz:
I implemented #760, due to release with 2.1 as well.

ClaudeMetz:
The basics of the interface are established, to be released with 2.1. I'll consider this main issue closed, of course many specific integrations will need to be done, but they'll get their own issues.

PR #750, opened 2026-05-03
Machine upgrades mod support PoC

### LINKED ISSUE #749: Compatibility with Machine Upgrades
### Suggestion

There is a helper mod out there called "[Machine upgrades](https://mods.factorio.com/mod/machine-upgrades)". It's used by the mod "[Planet Rubia](https://mods.factorio.com/mod/rubia)" to add productivity bonuses to buildings from tech. Unfortunately the bonuses it gives do not get picked up by Factory Planner. Would it be possible for Factory Planner to be made aware of this mod and take its productivity bonuses into account?

See https://github.com/LoupAndSnoop/machine-upgrades/issues/2 for a matching issue.

Wesmania:
This allows machine upgrades added in "[Planet Rubia](https://mods.factorio.com/mod/rubia)" mod to be accounted for when planning. Fixes #749. This needs a matching PR https://github.com/LoupAndSnoop/machine-upgrades/pull/3 to work.

Right now the added bonuses show under "recipe bonuses". I'm not sure if this is what we want. Productivity bonus does not apply to all recipes, only to intermediate products, so maybe it should be under beacon bonuses?

### DIFF
--- a/modfiles/backend/data/Line.lua
+++ b/modfiles/backend/data/Line.lua
@@ -182,16 +182,26 @@ function Line:setup_beacon(player)
     end
 end
 
+local has_mupgrades = remote.interfaces["machine-upgrades-get-modifiers"]
 
 ---@return boolean uses_effects
 function Line:uses_beacon_effects(player)
     return self.machine.proto.effect_receiver.uses_beacon_effects
 end
 
-
 function Line:summarize_effects()
     local beacon_effects = (self.beacon) and self.beacon.total_effects or nil
     local merged_effects = util.effects.merge({self.machine.total_effects, beacon_effects})
+    if has_mupgrades then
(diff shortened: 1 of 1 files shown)


ClaudeMetz:
Hi, thanks for the contributions, sorry for the late reply.
I'm sorry but I don't want to accept this PR. Having specific mod compatibility hard-coded is something I'm avoiding, as it would clutter up the mod a lot and put quite a lot of maintenance burden on me. I know that sucks, but it has to be that way currently.

I do have a plan for a more generic approach to inter-mod compatibility, as seen in #332. It's a ways away from being ready, but I know it's an issue that needs to be addressed at some point.

Wesmania:
Understandable. I'll keep it for personal use then.

Issue #746, opened 2026-04-25
Crash when choosing fluid recipe


BloodStainedCrow:
### Problem Description

Trying to use FactoryPlanner with Ultracube I had an issue with crashes related to steam recipes. 

My mods are 
```
Base mod 2.0.76
Elevated Rails 2.0.76
AAI Loaders 0.2.11
Factorio Library 0.16.5
Factory Planner 2.0.50
Krastorio 2 Assets (Ultracube) 2.0.0
Ultracube: Age of Cube 0.7.0
```

My save:

[CUBE.zip](https://github.com/user-attachments/files/27087912/CUBE.zip)

Here is the crash-log:
```
The mod Factory Planner (2.0.50) caused a non-recoverable error.
Please report this error to the mod author.

Error while running event factoryplanner::on_gui_click (ID 1)
__factoryplanner__/backend/data/Fuel.lua:31: attempt to index local 'proto' (a nil value)
stack traceback:
	__factoryplanner__/backend/data/Fuel.lua:31: in function 'init'
	__factoryplanner__/backend/data/Machine.lua:70: in function 'normalize_fuel'
	__factoryplanner__/backend/data/Line.lua:95: in function 'change_machine_to_proto'
	__factoryplanner__/backend/data/Line.lua:144: in function 'change_machine_to_default'
	__factoryplanner__/ui/dialogs/recipe_dialog.lua:92: in function 'attempt_adding_line'
	__factoryplanner__/ui/dialogs/recipe_dialog.lua:371: in function 'handler'
	__factoryplanner__/ui/event_handler.lua:173: in function <__factoryplanner__/ui/event_handler.lua:121>
```

### Reproduction

I added a recipe with steam as an ingredient:

<img width="973" height="122" alt="Image" src="https://github.com/user-attachments/assets/58506c8c-d082-451b-af22-07e0fff5ebcf" />

When choosing the temperature/recipe:

<img width="490" height="758" alt="Image" src="https://github.com/user-attachments/assets/4e6d519f-ec60-4bbb-b7b8-e46c2478bb1a" />

I chose the boiler recipe at the bottom:

<img width="518" height="865" alt="Image" src="https://github.com/user-attachments/assets/95bd5a9a-4fc7-48c2-b2c1-ff9f2766d47a" />

And the mod caused a fatal error:

[...]

PR #764, opened 2026-06-24
Fix fuel category generation

vapopescu:
Fix for #746

When generating the fuel categories, while iterating the machine prototypes, there was an extra delete, causing some machine prototypes to be skipped. This delete is already handled by the `remove_prototype()` deeper in the loop.

@ClaudeMetz Does the generator run on mod update? If not, what should be done to fix existing save files?

### DIFF
--- a/modfiles/backend/handlers/generator.lua
+++ b/modfiles/backend/handlers/generator.lua
@@ -800,6 +800,7 @@ function generator.machines.generate()
         }
     end
 
+    ---@param machine FPMachinePrototype
     local function insert_machine(machine)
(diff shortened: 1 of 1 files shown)


vapopescu:
Now I see that `handle_configuration_change()` runs the prototyper (and generators) again, so it should be fixed on update for existing saves as well.

ClaudeMetz:
Nicely done, appreciate it! Will be released with the next update.

Edit: And yes, everything prototype-related is refreshed automatically, so no need for any migrations.

PR #765, opened 2026-06-26
Object dummy constructors and type fixes

vapopescu:
* Add dummy constructors
* Fix type warnings


ClaudeMetz:
Thanks for the contribution. This is certainly a lot to just drop on here. Can you explain what dummy constructors are / why you felt the need to add them?

vapopescu:
At first I encountered the bug there the planner crashes if you paste modules into an empty beacon slot (fixed in 2.1, but not 2.0). After seeing what caused it and pulling the fix in my 2.0 fork, I considered that adding a complete simplified prototype to every object instance that was missing one should help prevent those kind of bugs in the future. So I also created a convenient constructor for those kind of objects `initDummy`.

But then I went into the rabbit hole of actually checking whether the methods that used the prototypes actually checked whether they are packed or not. And that caused the merge request to become quite big, admittedly. And I only went through the code of `Object` and the derived classes.

ClaudeMetz:
Okay I see. I've, with 2.1, just gotten a workable language server up and running again, and I've been doing through all the warnings to patch them up. I feel like it's important for me to go through this myself as well, so I can make choices on any architectural changes that the type issues throw up.

The dummy thing is indeed quite a mess currently, and your approach seems like a way more  proper way to do it. From a cursory look it all seems sensible, if a bit wordy, line-count wise. I'll try to review this soon.

If you're up for some bonus credits, there's the situation of SimpleItem, which isn't a proper object, but still needs to be paste-on-able. It goes as far as [this](https://github.com/ClaudeMetz/FactoryPlanner/blob/master/modfiles/ui/main/production_handler.lua#L281) abomination, and there's a few other places where SimpleItem needs custom handling because it's not an object. I don't want it to be an actual object because it's ephemeral data, fully generated by the solver, but maybe it should get some structure so the same methods can be called on it as can be on objects. I haven't thought about it too much yet, but if that's something you feel like having a crack at, feel free.

Edit: Also, can you drop the extension recommendation thing? I don't really think that's necessary.

vapopescu:
I'll take a look on the SimpleItem situation on a separate branch.

ClaudeMetz (commented):
Had a first look, and it's looking pretty good. I appreciate the type fixes. Two things:

1. Tiny thing: The method's called initDummy, but the project uses snake_case for function names.
2. I don't like the initDummy functions basically copying the majority of the normal init functions. I'd prefer if the normal init function had a flag or similar to produce a dummy version of itself. Maybe the prototype argument could be optional and that's the trigger.

vapopescu:
My bad for the casing. As for you second suggestion, the parent argument would have to be moved to be first. I suppose there's no issue with that, right?

ClaudeMetz:
Sure that's no problem. I went through the PR properly, and did a few adjustements. As I went over stuff, I just ran into tons of small things I didn't quite agree with. I cleaned some up, but there's still some stuff I'm unsure about. The fact that this rolls many changes into one makes it hard to review unfortunately.

I'll give it a couple of days and have another look then, but yeah it's tricky. And to be clear I'm not saying this is you doing things wrong either, it's just that I'm very particular and protective about how exactly this project is done. It's important because if I don't enjoy working on it, it loses its whole point. So if I accept stuff that will annoy me in the future, it's a bad idea. 

Anyways, please let me know if you disagree with any of the changes or whatever else comes to mind.

modfiles/backend/data/Beacon.lua
```diff
@@ -20,10 +20,9 @@ script.register_metatable("Beacon", Beacon)
 ---@return Beacon
 local function init(parent, proto)
     local this_proto = proto or {
-        name="",
-        category="beacon",
```
vapopescu: Maybe keep a blank category like the other prototypes

modfiles/backend/data/Recipe.lua
```diff
@@ -26,15 +26,13 @@ script.register_metatable("Recipe", Recipe)
 ---@return Recipe
 local function init(parent, proto, production_type)
     local this_proto = proto or {
-            name="",
-            category="recipe",
```
vapopescu: same here

vapopescu (commented):
Alternatively, you could also make the `category` field optional in the `FPPackedProrotype` class specification.

Other than these, the changes look reasonable. I should have payed more attention to the commenting style to begin with.

Also, not sure why you prefer nesting ifs rather than exiting early, but it's not that important anyway.

And I agree that having fun working on the project is the most important (I also have fun using it). So, keep it up!

vapopescu:
I also made the changes regarding `SimpleItem`. You can preview them here [vapopescu/FactoryPlanner#1](https://github.com/vapopescu/FactoryPlanner/pull/1)

ClaudeMetz:
Yeah category should definitely be defined as optional, only some prototype types have a category. I'll fix that. The existing types aren't to be relied on as they've not been kept up with in years.

I do think early exiting is reasonable a lot of the time, and left some of those in. It's just that the validation logic is very fragile and just changing it for formatting sake is not worth it imo.

ClaudeMetz:
> I also made the changes regarding `SimpleItem`. You can preview them here [vapopescu/FactoryPlanner#1](https://github.com/vapopescu/FactoryPlanner/pull/1)

Alright sweet. Feel free to PR it against this repo.

I'll merge this here PR now, thanks for the patience, I know I can be difficult.

PR #766, opened 2026-06-30
fix bug where free variables weren't being scaled correctly to max value in row

scottmsul:
Therenas had noticed a bug in the matrix solver where a table entry was being set to a variable that didn't exist. This wasn't breaking the matrix solver entirely, as it only affected the recently added scaling logic from [PR 740](https://github.com/ClaudeMetz/FactoryPlanner/pull/740). However, it was preventing free variables from scaling to the max value in the row, which was one of the intended behaviors from PR 740. This new PR fixes this.


ClaudeMetz:
Thanks as always! Seems this one just works, not sure what kind of issue you're having on master. If you let me know the exact issue I can probably help out.

PR #767, opened 2026-07-02
Cast `SimpleItem` to `TLProduct` on copy/paste

vapopescu:
Some things to consider:
* Class field naming regarding amounts (required and satisfied) is inconsistent (`SimpleItem` - `amount`, `satisfied_amount`; `Fuel` - `amount`, `satisfied_amount`; `TLProduct` - `reqired_amount`, `amount`). Should consider renaming in the future.
* An error is being raised when trying to copy an item with a fluid prototype, but unset temperature. I did this because setting the temperature on a top-level product is currently not supported, and trying to add a recipe for it crashes (bug in the "add_recipe" modal when neither `recipe_id` or `fuel_id` is defined)
* Non-English locales need to be updated separately.

### DIFF
--- a/modfiles/backend/data/DistrictItem.lua
+++ b/modfiles/backend/data/DistrictItem.lua
@@ -10,6 +10,8 @@ local Object = require("backend.data.Object")
 ---@field proto FPItemPrototype
 ---@field production DistrictItemData
 ---@field consumption DistrictItemData
+---@field overall "production" | "consumption"
+---@field abs_diff number
 local DistrictItem = Object.methods()
 DistrictItem.__index = DistrictItem
 script.register_metatable("DistrictItem", DistrictItem)
--- a/modfiles/backend/data/TLProduct.lua
+++ b/modfiles/backend/data/TLProduct.lua
@@ -4,7 +4,7 @@ local Object = require("backend.data.Object")
 
 ---@class TLProduct: Object, ObjectMethods
 ---@field class "TLProduct"
----@field parent Factory
+---@field parent Factory?
 ---@field proto FPItemPrototype | FPPackedPrototype
 ---@field defined_by ProductDefinedBy
 ---@field required_amount number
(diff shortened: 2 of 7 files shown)


ClaudeMetz:
I haven't looked at the implementation yet, but top level products don't support temperature picking because they are always a specific temperature. So it'd be `water-25` or whatever, not `water` with a configured temperature. Not letting you copy an unconfigured ingredient thus makes sense, although I'd probably prefer that pasting on top level products is the part that errors, so you can still paste on line ingredients. Hope that clears some things up, let me know if there's any questions. The whole temperature handling is complicated, but should be feature-complete.

The point about naming is very valid, I'll have a look at that as well. It's something that's a bit scary because it touches so many things.

vapopescu:
> The point about naming is very valid, I'll have a look at that as well. It's something that's a bit scary because it touches so many things.

It's very scary indeed. A good pre-requisite would be correctly typing all instances of `SimpleItem`, `TLProduct` and `Fuel`. After that, VSCode should be able to find all references of the fields.

vapopescu:
I need to think some more on how to paste fluid/fuel settings on the line ingredients. Ideally, pasting a compatible fuel should also change the recipe to use that fuel.

I will open a fix for pasting fluids without temperature on top level in a separate branch.

ClaudeMetz:
Alrighty thanks. I made a note about the amount naming issue so I can have a look at some point in the future. I wouldn't recommend you have a go at implementing that btw, since reviewing that would be just as much work as me doing it from scratch I'd think.

vapopescu:
@ClaudeMetz Yeah, this is what I thought. I will let you to it.

As for copy/pasting ingredient temperature settings, what do you thing about passing handler tags from `production_handler.handle_item_click() -> clipboard.paste() -> Line:paste()`? In particular `item_category` and `item_index`.

ClaudeMetz:
I'm not sure what you mean by that, you'll have to explain more or show me in the code :)

PR #769, opened 2026-07-02
Add error on fluid paste on top-level with unset temperature

vapopescu:


ClaudeMetz (commented):
Why did you use the messages system instead of the floating text the other pasting errors use?

vapopescu:
For no good reason. Will change it.

ClaudeMetz (commented):
Alright, thanks, will merge.

PR #771, opened 2026-07-02
Allow pasting SimpleItems on Fuel

vapopescu:
Automatically convert items in clipboard to fuel when pasted.

Fluid fuels tested with `bobpower 3.0.0` installed.

### DIFF
--- a/modfiles/backend/data/Fuel.lua
+++ b/modfiles/backend/data/Fuel.lua
@@ -108,6 +108,9 @@ function Fuel:paste(object)
             fuel:build_temperature_data()
             fuel.temperature = object.proto.temperature
         end
+    else
+        -- Only allow pasting items and fuels
+        return false, "incompatible_class"
     end
 
     if fuel then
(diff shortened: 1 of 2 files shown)


modfiles/backend/data/Fuel.lua
```diff
@@ -90,27 +90,42 @@ end
...
+
+    local fuel = nil
+    if object.class == "Fuel" then
+        fuel = object
+    elseif object.class == "SimpleItem" then
+        -- Try to convert SimpleItem to Fuel
+        local proto = prototyper.util.find("fuels", object.proto.base_name or object.proto.name, burner.combined_category)  --[[@as FPFuelPrototype?]]
+        if proto then
+            fuel = init(nil, proto)
+            fuel:build_temperature_data()
```
vapopescu: Temperature data is built on init

ClaudeMetz (commented):
Yep this is nice, thanks. I think this is minor enough to not get a changelog entry, as it's something people would have expected to work anyways.

PR #773, opened 2026-07-03
Fix line pasting into floors

vapopescu:
- Fix a bug which would morph the current floor back into a line when failing to paste on an empty floor. This in turn causes 2 issues:
  - Context corruption on save + reload (GUI is open on a deleted floor)
  - A second paste attempt would paste in a floor above.
- Improved the line pasting check for relevancy on the current floor.

### DIFF
--- a/modfiles/backend/data/Floor.lua
+++ b/modfiles/backend/data/Floor.lua
@@ -160,16 +160,48 @@ function Floor:check_product_compatibility(object)
     if self.level == 1 then return true end
 
     local relevant_line = (object.class == "Floor") and object.first or object
+
     -- The triple loop is crappy, but it's the simplest way to check
-    for _, product in pairs(relevant_line.recipe.proto.products) do
-        for line in self:iterator() do
-            for _, ingredient in pairs(line.ingredients) do
-                if ingredient.proto.type == product.type and ingredient.proto.name == product.name then
-                    return true
+    if relevant_line.recipe.production_type == "produce" then
+        for _, product in pairs(relevant_line.recipe.proto.products) do
+            for line in self:iterator() do
+                -- Check if pasted line produces an ingredient on a line on this floor
+                for _, ingredient in pairs(line.ingredients) do
+                    if ingredient.proto.type == product.type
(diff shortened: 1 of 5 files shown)


ClaudeMetz (commented):
Thanks for the PR. A couple things: I know it's annoying, but try to not bundle unrelated changes into one PR. The clipboard change and the robustness check don't really belong in here. No need to remove them from this PR, just wanted to mention it.

The dummy change is nice, I always wanted to make that less awkward. Could use a comment on why the true argument on :remove is there. 

The robustness check is a kind of thing I just don't like. You included it in a few places beforehand. I get wanting to prevent crashes, but it's just not a good idea in my opinion. If we as programmers can reason that a certain state should not be reachable, we shouldn't put a check for it in place. If the only way for it to matter is a bug in the code, I want it to crash, so we can fix the bug, and not hide it so it can cause problems elsewhere potentially. I don't think this is the only way to develop software, but it's the way I do it for this project (and it's the way Factorio itself does it btw).

The expanded compatibility logic is more correct I guess, yeah. I am worried about performance a bit because of the triple loops, but it's probably fine. I just don't like how involved this verification is, but I'd have to think about whether there is any better approach to figure this out.

vapopescu:
I had a feeling you wouldn't line the context validations at init. I'll remove them.

I'll also try to stop bundling multiple fixes int one.

> If we as programmers can reason that a certain state should not be reachable, we shouldn't put a check for it in place.

I disagree with this. Ideally this sound nice, but programmers are not perfect beings. We make mistakes.

But of course we can't put checks for everything everywhere. There has to be a middle ground, and I think type annotations is a nice way to do it without impact on runtime.

ClaudeMetz:
Yes, type hints are obviously a good idea. And yes we do make mistakes, but I want to know about them so I can fix them, not 'hide' them! But like I said I get other perspectives, it's just not something I want to do for this project.

ClaudeMetz:
I'll merge this, but I do want to think about a better approach to the compatibility check in the future. Thanks as always!

Issue #93, opened 2023-07-25
Define product amounts by machine


ClaudeMetz:
I have not done this for a long time since it has quite a few issues. People do ask for it regularly though, so I think I should do it, even if it’s not quite as intuitive as I’d like it to be.

The fundamental issue is that you can’t define a product only by machine count. You need the exact recipe, machine prototype, and module effects to be able to convert a machine count to a product amount. So, defining all this in the product dialog is just not feasible.

The solution I came to maybe like is to have the user enter a machine count in the product dialog (which disables the other options of course). Then, at first the amount will be determined to be 0. Only when they add a line to satisfy that product (it’ll always pick the first one it finds) will it be able to determine a product amount for it.

I’m not quite sure how it’d work technically, but the idea is to, for any machine amount defined product, find the first recipe that produces it and set a hard machine limit on it corresponding to the user-defined number. The solver will then produce as much as the user wants, taking into account modules etc. This number then just needs to be fed back to the top level product in some way.

This solution is actually less crappy than I imagined, so I should do it sometime. Might even need no solver changes at all, could likely be done at the interface level, which is good because then I don’t have to update the matrix solver.

I think I might want to remove the existing machine limit feature if I ever implement this. It was my initial solution for it, so this would replace it in a sense. Plus, it doesn't work with the matrix solver, which has always been awkward. This new solution would work with the matrix solver I'm pretty sure.

ClaudeMetz:
This is gonna be done. It's necessary to remove machine limits as they are non-linear for solvers, precluding features we want to have. The exact implementation might turn out complicated, but this is very necessary.

ClaudeMetz:
Implemented as #902 for the next release.

PR #778, opened 2026-07-12
Add new simplex matrix solver

vapopescu:
Pros:
* No need to configure free variables
* Supports "independent" floors (limit exporting/importing intermediates to/from adjacent floors)
* Supports machine and item limits

Cons:
* It finds *a* solution, but it may not be the one you want
* Defining a good objective function is hard
* There's currently no way to indicate what went wrong when no solution is found
* In order to truly optimize stuff, there needs to be a cost associated with each item


ClaudeMetz:
Alrighty cool. I'll still have to think about the UX of it all, but appreciate you already integrating it and stuff. I won't be able to look into this too much in the short term, as I want to focus on a few other things first. But I am interested in it as I said on Discord. @scottmsul feel free to have a look at this as well of course :)

Is there any context on the performance of this? Did you optimize it particularly, or is there still room for that? One detail I wonder about too, if performance won't be great, is whether it's possible to split up the calculations into chunks, so they could potentially be run over multiple ticks. That would require an estimation of how computationally expensive a certain solve is too, so it can be split up sensibly.

vapopescu:
I didn't notice any issues with performance, but the factories I tested were not very large. I'll try planning one with all Nauvis sciences and see how it goes.

ClaudeMetz:
Nice, that's a good sign already though I'd say, a previous implementation was noticeably slow even on small factories I think.

vapopescu:
I tried a factory plan for all sciences in vanilla Space Age (with scrap recycling disabled):

Matrix solver: 98.759000ms
Simplex solver: 3607.398900ms

I reckon it needs more time in the cooker :D

Factory used:
```
eNrtnVFv3LgRx79KsM9mGtuXIgjQh/ahT1egaPt2CBYURWmZUKKOouxs43z3DiWtrF2vXSe31FKrP3BnxJQsiiL548xwOPNtJb9Wxrp1YdJautXHb6uE13L1cXXz9vrt9fXqapWYRKvEcrul0tu3796+o0Kp5R13MmWWK12Pb8+4cMZuK83LUtr+yp/pgpViK/RQtHt2Ze7bst2Tf2+4Vm47vquuuJCM56PX+r6rSEmq/bdvK6F57d/j71319FclL/wf/KaK/C/KyeJPvHGm4E6ZktVCyZKeSQ/+8unNX4cLb/oLb/wFegaVWvV1XRt9J+2aHq3u6JnONnK4llkp1/75VPs3eq1EN7KyqnTrmq6XeV9cGuffdEUPraxJG/8oauY6MaXq7uiL95vzn1//2RV3f+eM76G+Zc+0h+4U1DW5aTvMvxmVpNzxtdtWsi+q/WdVRaVVpmTaNYneIZWZKmW6Tvyf8sI0pWt77vdGWSruSz5ef7963Stqk6vaKRHtCxaKBhsN7WhfUGxkoQTX0b7gbjRHPAob53s53j7u+BbtFJGOa93YPOJpzHOrRKNdYyOeKbRmCmcJ2XkpY2aisFuTyzLiN6TfaVRuVFNM/4qfrlbOVOtMG2P9Kw2CR1twtSLBSGq6kf5FDzwQTvbvuTl2z69U0klLyr/z44V/dSU/8jXGbe8e+EzrHyG+u70r8dXRN6uk5TS3ZCvL+I9l6B0/ZlzXJAftS0V0q5Cla4W163fvvJgkNr5Jo5b8oy96aeh5+dHtd+hwsT5oWl/FM23rBcr1YV2lsQXXB4/qblbPPSsz1Lq1VoVygxhI30nLdS88D01sS/8tfRO6O/Z7ubv+wnLqhUPW/SW73f8O41sOv0RfVfAPsZsPH75/8gMikVyYcvwB/taVPG1gsrswrq0rnPC9b07ecXUlSR063mPttXN31U3XVcIUhfS/r1Zj3P44dhKiqWUyzxnNCiZ45THAulKSaO699jdHBD3TkDGIjt3Cmpq0PTZ8lZgYlTV+uRmtQ/73Yz1qMt183W9sZszh0PWPO1r59x+aVdSPpxyP9FlKR2sglQtZ1ybU6FvRNzLWfJGOhiATRtNT2PX7d0FH5aHseGR5PLgFi2Q0i+R7LJJLXSQFt4kpWaaSea6GtCSIDS+SwyXQ2JyTJD7DRa5snCUlydX7DRoXB1jrLpBqv4BqS6XagQRkNEm+k0hbG+NtMB+el7X6X2GFuPARDbvDnOHzaYyfMEbRZ43YF6MSglKwlYJZEJj+mMAEEkFeAnuWwp6gzGl5cfPet6QwpeIkA92+B3ggAgFDwNCqJqEnZd6cyjJrCraDRCAijRlErbBNKplRXr88J5K8M2ft7PYBcAKcAKdo4ER1tv5zNGednOM+Xbf5T9TJGlvytvpxnxVSu6cqJ1hz7gEM1iyWNaEEnx3Kkk4fu45HH4PwA+EHQIoPSD0e2kNmRjfdUS5ZcdueLQysoI1rJUCddc8MgAKgAKj4AEU1mdoZy3ZmZnYn28k7OzWtaorq85OTagmvveamG5WCN9DQwJtYeLOvS4E34A14A96E480xTWxUZoTk5RwxZLKs3hgrmefRwUGSg0tzpNGJDzIOw+Fx13LiPj+JFz3WHqw9WHsuw5v+/4SuCWIg7OvUW9dvY2yMLtogJ4O58Po9Dl2DaTh0DXl6v8GNn/ec4KRcuDAQB3wCiUAikAgkOkIiU/qxGI5EWuUbN4l3KzgEDoFD8+PQTnUK6d86vX5G12WR+LOUrP+7w84Ulmfe8ZXdK7fprE4PQzzbLdCEWDJAUyRoGqgRiE739B1tcCQN4dpf8i8Dd2Lhzi24s1TujM03oQQiye+2O70M9mugCdoa0PQah3whmqIhVc3MMmbo69WyhwXhJ/pRDB1sucBJ73gpqM1CWdEox7pcbeGCVIX1PBoyze313HNtAmQAGUBmAi9Xa0qWS27Z/UZKDcaAMWAMGHPa/XbSLCqQBWQBWUCW0yZyMRVNayZ4Qg0GYAAYAAaAOek++G6ziJEMI74EDBA+7IRf7e9QBU6q8pqd8YcI80phhxxEWqbBRkh2PHDkPOQc+OKANCDNrAIgtPabWcImUTkrVOk3vVOrtD526r0NgQ7gwOkYwIkFOI9aV/u/D7DCUinnyaDFhVmZIPIBz60SjaYO9EF5woc9QDZnZHPGYojFcMLFsPJAqUzKZJ7P+/QN+Aa+gW/g2zNzqEs8RpDIdPMV8hv4Br6Bb/PnG4AGoAFoANrFAG3bFPyLYVQsZF2fYft54lOQcHoBe8CeONjzWWq9pdUa9AF9QB/Q50z70r0I5KycZd6hvV1TZ+4P9buj15H647H/22Wopi5D7y+n9/tJX3voneiIkV8BmmLqAYATRjhhBOEiWtUGgAFgABgAJpj20nr7dD/hSAtH2raZQ9j1idKHFUY7WbLu3D4NAhoFQZekjOZU2p4nG40WxJqHeQ0LVDQLlGvKvPZUCJoHoyePj3fmPV7BHXAH3AF3WmmE20SlwchTNzprLAlYXKg0eDKwn0q/Ax6BR+DRuYOZEYbCZdp5PYVOE+bDbk0uy+P5LR4jCw23IcdFNAj6AAQtFUEjJak7lqP5HQ+WINU/+2VV7CQogi4G2QfgmQV4+niuQA/QA/QAPdPujw72IGPlPI/tvRTsbMNtilhnoA6oExd1hHnySoitCN6AN+BNGN5wLZS7QAEHxAFxQJwYw0ePd55YLrf1TB1QvXPp58458gl3sI0eFXduwJ3Fc8cbebsfG+Pg8Q6P934+cCFnHzP69V5eWJMgC2NNiiHLdkj39rDrkNRSOC/Bkxxf8rb+cY8dT0cF0kD6BWkmd6RonQu8sl07aQ0p3MI29Wa22eL8yx+eKH+mRQAOgAPgTA0c81Wl8vy0OYVSBdgANoBN/EdkYpBuwBvwBryZAW8mMCk3TvmWwKgMOMGoDDj9nDB0gKbb06Ap01s/na1JjGOZ5YUEmAAmgAlgCpn6ijvSzraXEU8FgQwQyAAImhuChh1rWeZeXGhK5YIdJm4Sqom3zUBsJ/AIItHCdLVfTqOrhUfVmeHE0zteCumN9tDWgCagKZI9tbqVGKruMoLvRu0tjxm63BnaehHnklt2v5FSY7JCRICIAAD9+Db4jyPo0cYRCDobye+2zCgd3qDrbbaC6xftucBOLNi5BXZmjZ0J/G8eOQMXHBAKghEINZVmZrnSwAySdgIsAMvpD4zXTkGKgRQD2AA20zjLPIZ3AHIg0QAygMzpIHNkObxszjx0WDUFz0vpLtsxGOABeOZ9hnPYFYIFGboX6AQ6RXSI84yOwScJ4gXPYLAJbIJnMJwN4RmMGQrPYHgGQ0QAgBblolf4CFncIkYW+AQ+gU+TCUi1IxCwe67hpoe9JeAFeJneelspYkAXhI/0mZpmcs7/2z0FQAKQACQA6XTyTqas5LYAZAAZQAaQCXUcPLey5CnoArqALqDLOfz1tMmVPx8FgzIMyqAT6DSVguUsL+vKWMcSqWcZp/PotvpwaGER++sQfwCYaE9+l7W0rs0KB1kGcAFcAJcF+Aue5JwBHAahP4FAl2Dd4Y0zBUdEP/AJfAKfICFBQgKBQKCl7K7vGKTNPUtlWbe5xalW4bExGYmudkXCVHQnFV7jwBfoBDotPYyXkHXt8RQyp+a0OX5/Kq/mkuJ7gU/gUwQnMk5kXdqHSyCC3XPXCk3ITg5IITs5hKinAApOHg8XElC0bAqW8zp8Zi3ACDACjC5zR26I0KesaC41Q/kSYzdDtwOp5rU31xmkmeCJDm0Of5XtG1tzABAAtIitOVJs2lNnCQ+mvkFnA4igs8HK/YdPyKp843xGdR9UXXx56mAYwuA0VBpHGvcHY3NO1AK0kM4d0Dqz+raR/G47NZCGSgEkAAlAApCO2bQ9kx4dmyagkrBNKiehkm+alRldPLQtvdhoAAlAApDi3Irr96a8zQabcWAVbOFgFTbjcBAF/AF/ICsdzf+RWDXP4AGd5EOCXtbYkrfVj/u1kBqBA2Li0Q14tNi4bUVliDaWtciZI2wSRaqYKr1Ql1rVZkwadVrCax9t1+gnR2bAG8g/4M30vkgDcYThGsABcAAcACe8wad2yDAN7AA7wE40ro7joG0gE8gEMoFME1h8fIhIYyU7boadbzIQcAacAWfisyzvgANbD5AD5AA5595c33cBgvMP+AQ+gU+T2KI7p0MoXyANSAPSTKF8PSIH6hegA+gAOtP49gwnRWdInaopqs9d3rYntGlj6IM28FwGbWIRcdoD6t3POdLGZFm98Sqhx87BWfeDS3OEDn28J2sM/fc/Zrs9pg==
```

scottmsul:
So I haven't looked at the code yet, but I downloaded the `recursive_solver` branch to test it in-game. One of the standard matrix examples I like is producing uranium fuel cells with both Kovarex and Uranium Processing, since it can recycle any uranium byproducts. When I tried this with Simplex, it seemed to produce strange results. For context I'll provide all three (traditional, square matrix, simplex).

Traditional (note the 1/min u-235 byproduct):
image

Square matrix (recycles the u-235 byproduct):
image

Simplex:
image

I'm not sure what's going on here with the Simplex solver. It's not using Kovarex. It's also not eliminating the U-238 byproduct, and is using way more uranium ore and power inputs than the matrix solution. Is this a bug?

vapopescu:
This is a bug in the data collection stage. I didn't handle catalyst recipes correctly. And as you can see, I also didn't handle them correctly in the UI update stage

scottmsul:
So the Simplex solver seems fixed now for all three (fuel cell + kovarex + uranium processing), but when I do just fuel cell and kovarex, Simplex gives zero buildings for kovarex:

Traditional and Square Matrix (same result):
image

Simplex:
image

I'm guessing this is "correct" in that it's technically minimizing the cost function? Perhaps the cost function weighs the same for u-235 and u-238, so without kovarex the inputs are only 120 total uranium, as opposed to 132 total uranium for traditional/square matrix.

scottmsul:
Also how does the cost function work exactly? Earlier you said it was the total number of inputs + total number of outputs. As another test, I gave the solver two ways to make copper cables:
1. molten copper (foundry w/ 4x prod 3) -> copper plate (em plant w/ 5x prod 3)
2. casting copper cable (foundry w/ no modules)

The first method, because of all the prod modules, uses only 316 molten copper/min:
image

The second method uses 400 molten copper/min, but is preferred by the Simplex solver:
image

When I add speed beacons to the EM route, it prefers the first method:
image

Is it adding the number of buildings and/or the energy usage into the cost function? If it's adding energy that might be undesirable, only because the units are so high on them and would dominate everything else (MW = millions of units).

vapopescu:
That is correct. Because both U-238 and U-235 are seen as intermediate products, they have a cost associated to importing them to the floor. In the eyes of the algorithm, importing 180 items is cheaper than importing 192 items. The earlier example works because importing pure ingredients (i.e. uranium ore) has no cost associated with it.

In order for the simplex method to work in this case, we would need either:

1. To add a line on the same floor which breaks the intermediate loop (i.e. uranium processing).
2. A way to let the user configure which intermediates can be imported without cost (and treat them as pure ingredients).
3. A pre-computed cost associated to each item in the game, so we can normalize the import/export costs. (e.g. importing 1 processing unit should be more expensive than importing 1 iron plate)

modfiles/backend/calculation/simplex_engine.lua
```diff
@@ -0,0 +1,597 @@
...
+---@field ingredients ItemList
+---@field machine_limit number?
+---@field machine_force_limit boolean?
+---@field fuel_ratio number?  how much of an ingredient is for fuel (treat as 1 if nil)
+
+
+---@TODO: Move this to a better place. Maybe let the user configure it
+-- The objective function is maximized, so positive values indicate a score,
+-- and negative values indicate a cost
+local objective_vector = {
```
vapopescu, line 26: All objective coefficients are here. As you can see, the solver does not optimize for ingredients, as they are free

ClaudeMetz:
Having a way to determine a cost for every item is something I think this'll need at least in some way, else it just gets silly too often. What exactly that looks like though is tough as we all know.

One basic weighting should for sure be that fluids are 100 times cheaper than items, as that's sort of the conversion the game uses implicitly. Plus, any items of type 'entity' (which is what I call stuff like ore on the ground, etc) should be free. Not sure if you take either of this into account yet but I wanted to bring it up.

vapopescu:
`entity` types are free, indeed. I'll also put a .01 modifier for fluid types then, should help a little.

scottmsul:
I'm wondering if we were to have three types of solvers, if it would make sense to have three exclusive buttons, instead of the initial implementation with a left/right binary switch and the square/simplex choice in the preferences. Basically I'm thinking if the solver buttons could be the same kind of UI buttons as the rate display (see attached image). I think this would make it easier to switch between simplex and square matrix, and would also allow different subfactories to independently use the square matrix or the simplex, instead of forcing all subfactories to use the same matrix solver type (since preferences are global).

image

vapopescu:
I would put radio buttons, but it's up to Therenas.

Similarly, the preferred solver option in preferences would have to accept multiple values instead of being two separate binary choices.

And each factory would need to store the solver in use and migrated accordingly.

scottmsul:
I think the preferred solver could be a dropdown with three options, given there are already other dropdown menus in the preferences.

I believe the solver choice of traditional/matrix is already being stored on a per-subfactory basis.

vapopescu:
The solver choice for factory is a boolean at the moment. It would need to be renamed an changed to an enum

scottmsul:
Alternatively maybe there could be a gear icon to the right of the traditional/matrix choice so it's next to the matrix option, and clicking on it could open a matrix solver menu (hovering over it could say something like "configure matrix solver"). Then inside this menu could be a checkbox for simplex, along with other options to affect the cost function if simplex is selected. But in general I do think it would be preferable from a UX perspective if different subfactories could be configured independently.

image

ClaudeMetz:
I appreciate the discussion but we'll figure out the UI once we're at the stage where we want to add the solver in the first place, and I don't think we're there yet.

vapopescu:
Switched from the standard simplex method to the revised simplex method with LU decomposition and eta factorization updates. Solver update time on the benchmark factory went down to 1094ms

ClaudeMetz:
That's a pretty crazy improvement, nice! Still potentially problematically slow, I'll have  a look myself soon.

vapopescu:
Keep in mind that compared the the RREF matrix solver, this solver calculates the item constraints per floor rather than per factory. Because of this, in my example factory the number of constraints is almost doubled. And since the complexity is around `O(m^2)`, I would expect a minimum slowdown of at least *4.

Another area I expect some savings from is to switch all matrices (in `SimplexTableau` and `LUDecomposition`) to use column major order. This should enable me to bypass some buffer copies.

vapopescu:
Column-major order for the simplex tableau matrix brought the time to 628ms. Not sure if I can go much lower then this.

ClaudeMetz:
I think that's pretty close to the matrix solver, great stuff. Any other changes you're wanting to do? 

I guess a better weighting algo is not strictly required to be able to release this. People can still use it and if it doesn't work for them, there'll still be the matrix solver. I never really wanted to offer three solvers since that's a big barrier of entry for users to understand. Having it in front of me now though, it's kind of a nice progression of complexity vs capability. The traditional solver for non-loop situations, then the simplex solver for loops not requiring config, but potentially not solving the way you want, then the matrix solver for full control but also most complexity to the user.

vapopescu:
> Any other changes you're wanting to do?

I'm thinking of a way to reduce the number of constraints that spawn when subfloors are used. But top-floor performance is pretty good already. If you're happy with it, you can merge.

> I guess a better weighting algo is not strictly required to be able to release this.

Indeed. And a cost calculator would require its own PR I think.

> People can still use it and if it doesn't work for them, there'll still be the matrix solver. [...] it's kind of a nice progression of complexity vs capability.

I'm glad you see it this way.

vapopescu:
After reducing the constraint number and fixing the initial basis heuristic, the solving time on the benchmark factory went down to 477ms.

I think I'm really done optimizing now. Feel free to review.

ClaudeMetz:
Really great stuff. Will get to review soon, but allow some time for this, there's a lot of other stuff going on.

vapopescu:
Sure, no worries. I know it's a lot, so take your time.

ClaudeMetz:
Sorry that I keep making changes to the codebase that interfere with your PRs. I'll get to the two smaller ones now, and will get to this one soon. I am pondering UX currently.

One topic I'm curious about: Is there any overlap between this and the existing matrix solver with regards to the core math implementations? You achieved great performance it seems, and in general it'd be nice to de-duplicate. Or if not dedupe, maybe apply some optimization ideas to the other solver as well if it makes sense. This is obviously out of scope for this PR, and you don't need to look into it if you don't care to, but I was wondering about it.

vapopescu:
> Sorry that I keep making changes to the codebase that interfere with your PRs

It's fine, don't worry. I can keep up :)

> Is there any overlap between this and the existing matrix solver with regards to the core math implementations?

No, not really. The only one I can think of that can be de-duped is the row-reduction step.

> [...] maybe apply some optimization ideas to the other solver as well if it makes sense.

Yeah, makes sense. I'll see if any of them apply to calculating the matrix RREF, and open anther PR if I find something.

vapopescu:
I have an idea for another optimization.

Considering that the factory only changes one line at a time (for the most part), caching the basis of the previous result and re-using it as much as possible should bring the initial solution very close to the optimum.

The downside is that the cached basis becomes a target for migrations (we could also just delete it).

ClaudeMetz:
Interesting idea, sounds like it could be a huge speedup. Migrating it is definitely a big no-no, but that's fine because the solvers are already re-run on every factory when mod versions change, so there'll always be a cached version I think.

ClaudeMetz:
Had a think about future input-based and per-floor-solver things, and had a cursory look at your PR, and I think you said this already but it looks all setup for that? After we get it merged, we should work towards making the other solvers use an internal approach and data structure that's more similar to your work. Then doing input-based and per-floor won't be too big a deal anymore I feel. We'll have to drop machine limits but I already have a better approach planned anyways (see #93). Should we get all that figured out, I'd be unbelievably happy, and it'll make FP probably have the best solver setup, as opposed to one of the worse ones currently.

Also, just to clarify, I won't really review the core solver for this PR. I am just not qualified for that, and I trust you to do a sensible job with it. So I'll just make sure I like all the parts where it interacts with the rest of the mod, and maybe some organizational things. I'll try to get a first review in this week, plus organizing the GUI differently like we eluded to.

vapopescu:
The input-based stuff is pretty much ready, but due to the cost function not being very precise, there could be some hiccups to sort out when it comes to that.

For the per-floor solver stuff, the Gaussian matrix solver would have to be changed to solve per-floor as well. Which in turn means free items have to be configured per floor as well. After that, it should be just a matter of the solvers agreeing on the line data and result formats.

I'm glad you like the structures. I was thinking of refactoring the existing solver result structures to use the same serializing approach I used instead of putting each item in their own box depending on their type. It should also make Durandal's work with the quality easier.

As for the solver input data structure, all that both matrix solvers really care about is how much stuff would be going in/out of a line if it had only one machine (+ machine limits for simplex, at least for now). If you could align the sequential solver to use this principle as well, we could simplify things quite a lot.

ClaudeMetz:
That all sounds great, thanks for all you do. Let's get this merged first and we'll go from there!

modfiles/backend/calculation/simplex_engine.lua
```diff
@@ -0,0 +1,532 @@
...
+---@field floor_id ObjectID
+---@field products SimplexItemList
+---@field ingredients SimplexItemList
+---@field total_crafts number?
+---@field machine_limit number?
+---@field machine_force_limit boolean?
+---@field fuel_ratio number?  how much of an ingredient is for fuel (treat as 1 if nil)
+
+
+---@TODO: Move this to a better place. Maybe let the user configure it
```
ClaudeMetz: I'd hold off on user configuration for now. If the user wants more control, they can use the gaussian solver. We'll see how that works out though.

modfiles/backend/calculation/matrix_engine.lua
```diff
@@ -727,7 +727,7 @@ function matrix_engine.get_line_aggregate(line_data, player_index, floor_id, mac
     local speed_multiplier = 1 + (total_effects.speed / MAGIC_NUMBERS.effect_precision)
     local energy = line_data.recipe_energy
     -- hacky workaround for recipes with zero energy - this really messes up the matrix
-    if energy==0 then energy=0.000001 end
+    if energy < MAGIC_NUMBERS.minimum_energy then energy = MAGIC_NUMBERS.minimum_energy end
```
ClaudeMetz:
```suggestion
    energy = math.max(energy, MAGIC_NUMBERS.minimum_energy)
```

modfiles/backend/calculation/simplex_engine.lua
```diff
@@ -0,0 +1,532 @@
...
+        fuel_ratio = fuel_ratio
+    }  ---@type LineMetadata
+end
+
+
+---@param key PrototypeKey
+---@param amount number?
+---@return SolverItem
+local function string_to_item(key, amount)
+    local split = string.find(key, "_", 1, true) or 0
```
ClaudeMetz: This is problematic, because it could collisde with real item names. I don't love the string concat solution in general, but I also don't love the tiered tables that the sequential solver uses. I guess this is okay, but I suggest using a different separator that can't be part of a prototype name, and also that isn't `|` so it doesn't collide with the temperature string thing.

modfiles/backend/calculation/solver.lua
```diff
@@ -138,6 +139,7 @@ end
 ---@param calculate_emissions boolean
 ---@return FloorData
 local function generate_floor_data(player, factory, floor, calculate_emissions)
+    local calculate_emissions = lib.globals.preferences(player).calculate_emissions
```
ClaudeMetz: Why is this here?

modfiles/ui/dialogs/machine_dialog.lua
```diff
@@ -307,8 +307,13 @@ local function open_machine_dialog(player, modal_data)
...
     -- Limit
     local factory = lib.context.get(player, "Factory")  ---@as Factory
+    local line = modal_data.line
+    local floor = line.parent
     -- Unavailable with matrix solver or special recipes
-    local limit_enabled = (not factory.matrix_solver_active and modal_data.line.recipe.proto.energy > 0)
+    -- Limited availability with simplex solver (top floor only)
+    local limit_enabled = (not factory.matrix_solver_active and line.recipe.proto.energy > 0)
+            or (factory.matrix_solver_active and lib.globals.preferences(player).use_simplex_solver and
+            (floor.level == 1 or floor.level == 2 and line == floor.first))
```
ClaudeMetz: I'm not sure if this is worth supporting if it's only for the top floor. That plus I am quite sure it'll get dropped anyways in the future, because solvers per subfloor will make it impossible, plus I have a better solution to the problem this solves in mind anyways.

modfiles/util/lib.lua
```diff
@@ -14,7 +14,10 @@ local _lib = {
     actions = require("util.actions"),
     effects = require("util.effects"),
     temperature = require("util.temperature"),
-    preferences = require("util.preferences")
+    preferences = require("util.preferences"),
+    table = require("util.table"),
+    matrix = require("util.matrix"),
+    math = require("util.math")
```
ClaudeMetz: I don't like these utils. They are only used by the new solver, so I think they belong with it. I get the idea of wanting to generalize so other things could use them later on, but I'd rather make them a lib when it comes to that, and not before.

Maybe it's time to have a dedicated solver_util file in the /calculation folder that has these kinds of things, plus what's now in solver.util. Maybe it can also be combined with the structures.lua file.

Have a go at adjusting it, or tell me if you want me to move things around.

ClaudeMetz (commented):
A few first notes. Code looks pretty good generally though, as yours usually does. I'll do the GUI adjustements in a bit.

modfiles/backend/calculation/simplex_engine.lua
vapopescu: Yeah, I agree. All the `string.sub` stuff was just so I could make it work. String split would be more elegant.

modfiles/backend/calculation/solver.lua
vapopescu: Leftover from a refactor which turned out to be pointless. I will revert it.

modfiles/ui/dialogs/machine_dialog.lua
vapopescu: Ok, cool. I thought of it more like an experiment to see that the solver can indeed handle limits on any variables.

modfiles/util/lib.lua
vapopescu: Sounds reasonable. I'll move just the utils for now and I'll see about the `structures.lua` when I do the refactoring I mentioned before.

ClaudeMetz: Alright thanks. The solver code that isn't your new one is quite messy anyways, so it doesn't need to be perfect for now. Especially with the big changes still coming.

ClaudeMetz:
Alright I changed it over to offer all three solvers at once. Let me know if that works for you.

Also let me know when you're done with your changes, I'll do another review pass but otherwise I think we're pretty close.

I had a super simple performance check, and while your solver is about 4x slower than the sequential one, it's about twice as fast as the gaussian one, so that's amazing work, well done.

vapopescu:
Yes, it looks good. Nice job as always!

You probably used a design with more floors, which as it stands is a huge advantage for Simplex, as it solves each floor at a time instead of all the factory at once. It also caches and re-uses the results for unchanged floors, which is pretty unfair :D (the cache is invalidated on refresh button and configuration change).

In my testing on the Space Age factory below (one big floor to make it apples to apples), Sequential solves in 8ms, Gaussian solves in 85ms and Simplex in 500ms.

Of course, the solver can yet be improved. But I think it's in a pretty good spot right now, considering both numerical stability and performance. So if you think it looks good, you can merge it.

```
eNrtnV1v2zgWhv9K4euwm6QToAiwF7MXezUDLHbmblAYlETJbChSw4+k3jb/fQ8l25G/su2OKUvxe5eQskiK4sOX1DmHX2fiS2Osn9emcMLP7r/OMu7E7H52+/7m/c1Ps6tZZjIlM8vtklI/vL9+f02JQolH7kXBLJfK9S8vee6NXTaKay3sOif+xop8mauXtNXNG/PUpq1v/WfgSvpl/yrX8FwwXvXq9bwuSQoq/o+vs1xxFyvyz658+pXmdfzBH7Ku/i69qP/Ggzc199Jo5nIpNN2Tbvzw6d3Pm4x3q4x3MSOWbNRjWzsn60aJL5REF1r5ZV5aIebxtlToV6pNpoJorNR+7ihfV6tkbXys4Ix+2FhThNzLR2rdPDNadleskrdb8fsv/+qSu995E3tm1aAjzaArc+qSyrQdFWtGKQX3fO6XjVgludim2BRZSlHM7r0NgupQiFJqUcyz+FNem6B922F/BmkpeZVyf/N89X1VVKaSzst8tBWsJb1j9EqPtoL5QtQy52q0FVy/zSN+C4OPvTzePu6wNtohIjxXKthqxMOYV1bmQflgRzxSaK7MvSVkV1qMmYm5XZpK6BHXkP6nt3IhQz18FT9dzbxp5qUyxsYqbfRGm3A1I0EkFF1If9ENt2fzXyilE0Ay1ucl499dyo+0tN+u7oZHWvYC6PXlXUosjp5HIyyncSNanRIfhKE63pdcOXE141GmiO5WdCdhc6F9q79urq+jBMoXsUm9lvy6SnrttYqa0G931ibT7TRtVcSRtq004ny3LG1szdXOrbqL5bF7lYZaN1eyln7dYBLDQYn5ShBvmtim/iZiE7ortnu5y39lqozCj3W/ZB+2n0P/kt0nsSoq+YNYv+sfnz/FFyITPDe6/wD+0aXsNzBbZ/RL6xIHrPftyTvONYKWOId7rM07d1fddl2Vm7oW8f/ZrI/SH8fO0SkgBXVmpQrGmgfhCT0sN4ruwm7urkEj0Ag0Ao3+l3BOwqRVmWrpI17urmYLo+pWhhkV4i1ialJC7bb6AKd2LgGtRkOrO9DqUmn12l7E5BZtmTT5gteZsNs9Z2zFSR+NCThliOvu3oI8/r/fIh28pT7xbrtB/eR+PeJdD9bh+dIR9xMQd6mIe2VLOIkYq43yQtPSsGniR7gb4lhS6pX02Ir222GvJzeNXkJmgUFg0JkZdPCj2eT0FeWLOlNSV2z1u93Oyy0vPeUDOoAOoHNm6ByxJAB2gB1gB9hJhZ3jNlYgD8gD8oA8yQwBjpifgjvgDrgD7iTbYT5ilw/ugDvgDriTijvHHJbAHXAH3AF3khkNHfXlBHlAHpAH5Em2w8NdHItMmSdWCO3az1tUZh5hkdigR9rOjrplDGx8wCawCWzasj72wjJRVYxGBct5E4c/61Jdw590a5Y8PU/YIw3pS6RDl7DgIqk3T2WC1teZNKUKX7YbWxpTpDG7pn485ftIj0X7UDNKz4Vzxr4lj0j4G8HfCJPkFAW8zYxmpcymORvCxQguRpD+oNrO8nxbARm1v1+YRm0tTAzz8xHRJ9Lqq9G/0Ig3AfYYLZIyp+XF7V1sSW205LTK+3AH8GBhBwwBQzMqsw38RyPViymu7LrtImJNGazmbfH9PquFwlfgMbHmFqy5cNakUjtrlGWd5LkZVPKsXRgOKZ6Y57xdfoP2GR+PPoBHFxvdwShZsLgzyrrjIv7y9k98BUI9tATanHWx1WXHmoQNHmwugy6D0qW0pmZrLZJI/fSlDrXChkIwI1Xc+oHygfKB8gGbeubzHSxan+tVqGHmRMNt6x2QGFH9UglXqWOwg0/gE/g0LT5RScZ5Y9n6CxZ7FO3gndz2dBPq5vPesUEZd3HHWgVZgDfYmQZvxsKb7T1k8Aa8AW/Am9Ovv/I81EFxb+zb9r7+dkHeHNiMBnBGuxkd4ojnDc9lOubsnmcFTzJIH3iSgUQHSGR0fBfTkUjJauEH+f4FDoFD4ND0OLQ+ajOl4fPw53l+/8qMPUm/6LaEviEiDdxSgaaxHUXcLqNSgUnwx+VaH+HcYaAJqglo+gHVtAFFIj490XO0ySn0XbZB4A4MgsCdcx9OFQd5Y4oDEekmxhxEIkIkIiz5wLcjY6jzFXmJYIlIa+Ab+Aa+TT7IMoAGoAFoANobAdoy1PzBMBebjeABsNcEXoCX0+NldQZAurC0YzETH6GYgrQBey7bGW4FIW/FJF3heGVlHhSVxRXz5mlX4xzMnyJ+TnwAzWeh1JIWgxC3ELeYYDDBnHyC2QAG8hbyFvQBfc4jb1sMOSIGxO3liFsfdOXaM4C5zWSRzMDaBVWGGJae57JI7oT2f7l9YDbCbITZ6Myz0YZHSf3Q9o5Cx6nn4A64AxVs2QZAxoppmueQ5pE6SpvCSqW2+27BbcHaSNzY6wNnwJkzHRz9NlZZCCUNzzEAaHLnrC6jNrAmM56VltfibX9rAHSwugJ0RhHBg3SJ0FUcq0FLnyzUWcioJN42A7vM4BF4BB7tdST3XuyFssAyDATCMgwEGlQRvRzTjkUYNpshcwCZE0YJ2p8OcbwGwAPwADxpwWO5VFA0AAvAArCc1kzHxpOZaWZ/AF6wRwPYADYp92iSf6w6M2148ch1Tp0K7AA7wM5Ydm26ZKgbLJ4AFoDlpNvBq/ACSQ1wJuDq+Q2H6UD4gE+jOwq1D463f5IObHJgkwP+jI0/ycET2RKPMxShZhV3sA8Ei8AisGinwZ5wwJ64wnd0bAVhqQW8nHqPmcZ35wcadOFolFb8P91dABvABrABbE7oci6t4LYGZAAZQAaQSQOZygrNC7AFbAFbwJYTWx1rJ6wXFnABXAAXwOW0EY8t164x1rNMKP/G/TMvIuoxaAPajJU2Lyb/0uZBeviD49s2GAQGDfrxiS6gwcgynszYZmALm7XtDExspoahj8DQpWLoJVBoIggtBH9cMiPVOEz8AB3Y9QE654aOrBY+MiH6XOcP6Y567dsbbwodia0xjn8FkACkcQBpI1KGBNLIlNEFAGn07zUQhD3pSKEXJ/UBOJTbUIhBOBSbZkVJmbsHgb7aaGgiAAlAOk9wdhM3dPGZDBTCZzJQ6GwuoHGPepLBTg+ccNHvsVooRBwcE2luQZrLPZTYtQGynBcilb85PSUVj1WPEZzj4uv6Oil9yujXurvUugjrQwgcYGda2GmDuleCW/a0AIAAIAAIABocQKAOqAPqgDoDnWFjrGCHt0AmsbkDzoAz4Mw01E1uGhrsLOdZsuM/1yqnK4qAcgOdA/6AP+DPhj8gD8gD8oA8g5CnUzxYY4E0IA1Ik3Ivp26M88ay9abOFEGTyYrVUketVljZRpzv9VvGncyZM2rvTB74RwAygMyAkHnRNcAMMAPMADOJMMMVAIOlE3AD3AyAm9bPAbyBoAFhQJg0goarXHowBowBY8CYNIzhdW205DHUDI3HEMnBvFSThI4pS7eIX8+aUDc70Sp2sqbIHHp4aaaZTTSRCfZ57M/PvHU13ptgShWwaIanLqab8Uw3bRAjTDBTmWDoLfj0/F/0SEuq
```

ClaudeMetz:
One thing that bothers me is the hack to set a tiny machine amount for recipes we don't want to calculate machine amounts for (like oil drilling). The hack itself is fine, but it shows <0.0001 to users which is confusing. You adjusted the margin, which is fine too, but before it didn't show the tiny machine number to users at least. It shows tiny power consumption and emissions for both, which isn't ideal either. So I think we should try and get rid of those somehow.

vapopescu:
The proper fix IMO is to calculate the energy value for infinite resource pseudo-recipes (oil). But I'll try to think of a nice workaround to achieve what you want.

Both matrix solvers have to have something there (preferably not too small, because it could be a source of errors), because recipe energy is in direct relation with machine amounts. And machine amounts are the main variables we solve for, so if there is no coefficient there, we can't get a solution.

The hack value is from the [official API](https://lua-api.factorio.com/latest/prototypes/RecipePrototype.html#energy_required) btw. You can lower it, but it will affect stability.

ClaudeMetz:
Nah the value is okay, we might just need to remove it again after solving. 

Calculating it for certain recipes is just not possible, because it relies on data we don't have. Like for oil, machine speed depends entirely on the percentage on the oil wells actually used, so there just isn't a number we could use.

ClaudeMetz (commented):
Yep that looks spot-on, thanks.

ClaudeMetz:
Okay I think that's all looking good now. I like the way you write code, aesthetically, btw. Pleasant to look at lol.

Anyways, yeah I think this is it. I'll do a changelog, merge, and do a release in the coming days, we'll see if there's anything else to squeeze in.

I'll also open an issue to discuss what the next steps with regards to solver improvements should be.

PR #779, opened 2026-07-14
Fix temperature configured logic

### LINKED ISSUE #777: Adding a line with a burner machine falsely triggers `warning_temperature_not_configured`
### Problem Description

Any recipe triggers this warning, even if the fuel is a solid item.

### Reproduction

1. New factory for 'Iron plate'
2. Add recipe for smelting 'Iron plate' (default stone furnace and coal)

vapopescu:
Fixes #777


ClaudeMetz:
Thanks! This was seemingly introduced since the last release, so I won't be adding a changlog entry for it.

PR #781, opened 2026-07-14
Fix fluid picker

vapopescu:
Fix crash when show unbarrelling recipes is enabled and picking 'Crude oil' to add a mining recipe


ClaudeMetz (commented):
Thanks, nicely spotted my oversight there.

Issue #782, opened 2026-07-15
Crash when turning on "ingredient satisfaction" option without first closing the planner window


liliohlili:
### Problem Description

This was observed on a fullpy map, but I don't think it's related to any mod interactions. 

### Reproduction

1. Create some FP factories
2. Turn on "ingredient satisfaction" in Preferences
3. Click on any other factory page, without first closing and reopening the FP window.

<img width="402" height="369" alt="Image" src="https://github.com/user-attachments/assets/c48509ed-3d55-4999-b8c3-da6c153e52e7" />

ClaudeMetz:
Thanks for the report, fixed for the next release.

PR #785, opened 2026-07-15
Prevent edge-case when converting satisfaction percentage

vapopescu:
Fixed an issue where if the ingredient satisfaction is enabled and the solver finds a solution there satisfaction is between 0.001 and the margin of error (1e-6), it will result in a crash when the UI is updated.

### DIFF
--- a/modfiles/backend/data/District.lua
+++ b/modfiles/backend/data/District.lua
@@ -105,8 +105,7 @@ function District:refresh()
     self.needs_refresh = false
     self.item_set:clear()
 
-    local filter = { archived = false, valid = true }  ---@type ObjectFilter
-    for factory in self:iterator(filter) do
+    for factory in self:iterator({archived=false, valid=true}) do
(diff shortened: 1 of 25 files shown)


ClaudeMetz:
Does this have anything to do with #782 ?

vapopescu:
Not sure. Considering the place of the crash, I suspect not. ~~Actually yes.~~
image

~~So fixes #782~~

The crash in the issue seems to be caused by passing nil as 1st argument to `util.gui.calculate_satisfaction()`

ClaudeMetz:
Alright just wanted to ask because the timing was pretty coincidental lol, all good.

ClaudeMetz:
Seems good, thanks. The other issues is indeed different, it's because I spread out refreshing of factories over multiple tricks when changing preferences so they don't lag, but obviously that's an issue if you select one that's not updated yet. I'll figure out a fix for that, the whole background refresh logic wasn't quite fully worked out anyways.

PR #786, opened 2026-07-15
Allow resource mining pseudo-recipes to use modules

vapopescu:
Fix issue where there are no modules to select for miners and pumpjacks


ClaudeMetz (commented):
Thanks once again.

Issue #787, opened 2026-07-15
Add surface restictions to resource pseudo-recipes


vapopescu:
### Suggestion

For example, you should be able to pump lava only on Vulcanus, mine scrap only on Fulgora, etc.

ClaudeMetz:
Where would I get that information from?

vapopescu:
It would seem that #536 is a prerequisite for this to work

ClaudeMetz:
I don't think it is? The data is only implicit on certain definitions. For the lava for example, you'd have to track that lava only exists on certain surfaces, and then when defining the lava pump recipe set the conditions appropriately. You'd have to do the same for all other kinds of custom recipes. It's not impossible, but there's a few issues. 

One, what if mods script lava onto certain planets, we'd be restricting too much in that case. Second, and harder, is that the surface conditions aren't a list of compatible surfaces, but a set of constraints. There could be lava on two different planets with wholly incompatible surface conditions, so it wouldn't be possible to find the union of them. You could duplicate the recipes I guess but that has other issues.

There's a similar thing where I want to track the availability of mining recipes for example, currently they are always marked as unlocked/researched. I could do that by tracking when the first machine that can run the recipe is researched, but there too scripted stuff could interfere. But yeah this one I think is actually doable at some point.

vapopescu:
Found something. The keys seem to be somewhat hardcoded with the planet name (beside Nauvis).

`game.planets["nauvis"].prototype.map_gen_settings.autoplace_controls = `
```
{ ["coal"] = [table *0x000002583a0aa7a0],  ["copper-ore"] = [table *0x000002583a0aa800],  ["crude-oil"] = [table *0x000002583a0a9900],  ["enemy-base"] = [table *0x000002583a0a9240],  ["iron-ore"] = [table *0x000002583a0a95a0],  ["nauvis_cliff"] = [table *0x000002583a0ab1c0],  ["rocks"] = [table *0x000002583a0ab100],  ["starting_area_moisture"] = [table *0x000002583a0aada0],  ["stone"] = [table *0x000002583a0ab6a0],  ["trees"] = [table *0x000002583a0aabc0],  ["uranium-ore"] = [table *0x000002583a0abfa0],  ["water"] = [table *0x000002583a0ab700], }
```

`game.planets["vulcanus"].prototype.map_gen_settings.autoplace_controls = `
```
[...]

PR #791, opened 2026-07-21
Block resource recipes based on location

vapopescu:
Implementation for #787.

There was an issue where offshore water was blocked on Gleba. This was because the pseudo-recipe was registered as `impostor-water-water` but Gleba has no `water` tiles.

To fix this, I changed the offshore fluid pseudo-recipes names to `impostor-<fluid.name>-tile` instead.


ClaudeMetz (commented):
This turned out pretty elegantly, simpler than I expected. Did a couple minor adjustements again. Only thing I don't like is referring to FPPrototypes in other prototypes. It works fine, I'd just prefer it so save a list of names. Esp. here where it's just a lookup, just doing a {name: true} table is preferred.

Otherwise looks good, do you want to do the change or should I? I want to get it out with the next release today ideally. The migration needs merging with the 2.1.7 one in that case too.

vapopescu:
I'll do it

vapopescu:
Done. I just noticed that you can grow yumako and jellynut on Nauvis, but we can fix it later.

ClaudeMetz:
Great stuff, thank you.

Issue #497, opened 2025-03-14
Ingredient satisfaction is broken still because it's linear


ClaudeMetz:
Fix it by doing two passes instead, one for counting products, one for setting satisfactions (top to bottom or bottom to top is a matter of taste). Will be slower but still fine probably.

-> Looking at it further, the current algo is correct for the traditional solver, and doing the above would be wrong for it. But the above is what the matrix solver needs. So I kinda need two different algos for this crap. Which I don't wanna do.

ClaudeMetz:
Fixed by #805 for the next release.

PR #805, opened 2026-08-02
Add satisfaction algorithm for matrix solver

vapopescu:
Fix for #497


ClaudeMetz:
Nicely done, thank you for picking this up.

I don't think I need to ask, but I assume this one works for your Simplex solver as well, correct?

vapopescu:
Yes. I pulled the changes now and it works.

PR #811, opened 2026-08-08
Add surface restricitons to planting pseudo-recipes

vapopescu:
Follow-up to #791. Now agricultural tower recipes are also restricted to the surfaces the plants can grow on.


ClaudeMetz (commented):
Nice one, thanks. I also took inspiration from your generator-local variable to dedupe work in a couple of other spots.

PR #812, opened 2026-08-08
Matrix RREF optimizations

vapopescu:
@scottmsul FYI

Some minor optimizations for the Gaussian matrix solver:
- Use absolute tolerance, since the matrix values are re-scaled in `matrix_engine.get_matrix()`, so the maximum cell value is always 1.
- Permute rows vector-by-vector instead of element-by-element
- Reduce the usage of `math.abs()`
- Normalize only the non-zero values of the pivot row
- Delete unused function (`lib.flib.shallow_copy()` already exists)


scottmsul:
All looks good to me! Thanks a ton for doing these optimizations.

ClaudeMetz:
Thanks for having a look Scott, and thanks vapo for getting into this!

PR #814, opened 2026-08-08
Solver structures refactor

vapopescu:
Part of ongoing effort to make it easier to support per-floor solving and quality:
- Make all serializations go through `structures.pack_item`
- Drop `SolverClass` and use `SolverMap` instead


ClaudeMetz:
Alright sorry for the huge wait, I hope it wasn't too annoying to any other improvements you are working on. I did a few additional cleanups. Thanks and let me know if you need anything else from me as always.

PR #815, opened 2026-08-08
Fix simplex not exporting intermediates

vapopescu:
Fix bug where setting an intermediate product as a factory product would have its demand not be satisfied.


vapopescu:
Fixes #834

ClaudeMetz:
Thanks!

PR #820, opened 2026-08-09
Fix pasting on fluids without set temperature

vapopescu:


ClaudeMetz:
Why the temperature_data check instead of temperatures?

vapopescu:
If the temperature is not set for an ingredient, it does not appear in the `temperatures` table.

ClaudeMetz (commented):
Sorry for taking so long to get to this, but what is the specific case this fixes? Pasting fluids with and without temperature seems to work fine.

vapopescu:
Currently, if you copy a fluid with temperature (Water 15) and try to paste it over a fluid with temperature unset (Water), it fails with error. With this fix, it copies the temperature successfully.

ClaudeMetz:
Ah seems I happened to only test pasting onto fuels, which avoids this obviously. Thanks.

Issue #819, opened 2026-08-09
Boiler pseudo-recipes can create fluids with temperatures that aren't used anywhere


vapopescu:
### Problem Description

A concrete example is Space Explaration's `se-energy-transmitter-chamber`. Because it is a boiler entity with unfiltered fluid boxes, FP generates fluid item prototypes with `goal_temperature = math.min(fluid_proto.max_temperature, input.maximum_temperature or math.huge)`, leading to absurd stuff like this:

<img width="369" height="299" alt="Image" src="https://github.com/user-attachments/assets/fd51c2a8-6627-4764-bc8c-48ea4c61f289" />

### Reproduction

1. Load provided save
2. Open FP
3. Add recipe for 'Astronomic catalogue'

[SE Test.zip](https://github.com/user-attachments/files/30878955/SE.Test.zip)

ClaudeMetz:
Thanks for the issue and the PR! Fixed for the next release.

PR #821, opened 2026-08-09
Fix boilers with special pipe connections

vapopescu:
Fix for #819.

When generating boiler pseudo-recipes, now only fluid-boxes that can connect to pipes are considered.


ClaudeMetz (commented):
Thanks. I adjusted it a bit to only check on input pipes since that's the relevant spot. This could drop legitimate machines that just have a different set of pipes with a separate category. I'm not aware of any mod that does this, so I'll leave it as is for now, but it could come up again later.

PR #822, opened 2026-08-10
Implement topological sorting

vapopescu:
There are a few more kinks to iron out and tweaks to make, but this should be it for the algorithm, more or less.

If you got the time, you can start designing the UI (I put a blank button before export floor).


vapopescu:
Implementation for #338

ClaudeMetz:
Okay so can you explain the purpose of this? I guess the purpose for the user is somewhat clear, but is there a specific purpose with relation to the simplex solver? Also, how does this deal with loops?

vapopescu:
It can be used for all solvers, but only affects the results of the sequential solver and the simplex solver. In both cases, it tries to minimize item loops:
- For the sequential solver, minimizing the loops leads to better ingredient satisfaction and fewer imports.
- For the simplex solver, minimizing the loops leads to fewer variables (better performance) and fewer items that can be imported/exported (which may lead to a better solution).

The way it works is that it treats the floor like a directed graph. The lines/subfloors are the nodes, and the items are the edges going from the nodes that produce them to the nodes that use them as ingredients.

The loops are detected while exploring the graph. When one is detected, it removes the item (edge) with the least priority in the loop based on an heuristic. The one I used goes something like this: remove the item that appears with the highest index in the corresponding line recipe where it is produced. In the case of ties, also compare the recipe ingredient indices.

We could also change the heuristic. Or try a different approach altogether, like pre-processing the graph in advance to remove the loops. I'm open to suggestions.

ClaudeMetz:
I see. How would you explain what this does to a user? Expanding on it a bit instead of just saying it sorts? Sort *how*?

vapopescu:
Uff, tough question.

I suppose the simplest way to explain it would be that it optimizes the supply chain order by putting products towards the top and ingredients towards the bottom.

ClaudeMetz:
So while I like this feature, I feel like it'll be something that needs iteration and fixing after release. I don't quite want to take this on currently, with all the solver stuff still on our plate. Is that okay with you? Keeping it around to have a proper look at in a while? I also want to look at using this to enable further features.

vapopescu:
I agree; not a priority.

ClaudeMetz:
Great thanks for being so flexible

PR #830, opened 2026-08-17
Add error messags to simplex solver

vapopescu:


ClaudeMetz (commented):
This is great context, but I don't think it's quite done in the right way. These warnings should persist, and not be a timed thing. I have plans to rework stuff surrounding this soon, as noted in #838. I'll fold this PR in with that once I get to it.

ClaudeMetz:
I won't fold this in with those changes after all, we'll do a whole ground-up rethink of solver-related warnings and states and errors when we do per-floor solvers. I'll close this PR as it won't be implemented in this fashion, but the idea is obviously good.

PR #831, opened 2026-08-17
Rework line/floor paste

vapopescu:
Pasting on a floor now inserts the clipboard into the floor instead of replacing it. When pasting on a line, now it also transforms it into a subfloor before pasting into it.


ClaudeMetz (commented):
Yeah this is nice, thanks.

Issue #832, opened 2026-08-18
Implement integration for SE's scripted resource generation


durandal42:
### Problem Description

I'd like to make a tiny plan to size a basic core mining operation in SpaceEx. FP thinks I can't do core mining at all, though.

### Reproduction

1. load attached save
2. open FP, and if it's not already selected, look at the very last plan in the list (named `iron ore, copper ore, stone, coal (core miner)`)
3. note that the last line of this plan has been greyed out (not visually obvious, since core fragments are already grey) due to "Location can't use this recipe" (see attached screenshot)
4. bonus round! remove the disabled fragment mining line, then re-add it by clicking `Core fragment` under Ingredients.
5. note that the bottom of the screen shows an error: `"Core seam (Mining)" is not compatible with the current location and has been disabled` (see attached screenshot)

<img width="1687" height="729" alt="Image" src="https://github.com/user-attachments/assets/a8922186-0482-4b32-855b-c454f8bea785" />

<img width="1172" height="88" alt="Image" src="https://github.com/user-attachments/assets/1f537b3f-70fc-4efd-abb3-cb326d34e497" />

[spex 014 - core mining.zip](https://github.com/user-attachments/files/31166491/spex.014.-.core.mining.zip)

durandal42:
Same thing trying to mine cryonite (from a planet that has some, but I don't see a way of telling FP that).

vapopescu:
All resources are fixed in #833. This feature was intended only for Space Age and modded planets. It doesn't work with Space Exploration (planet resources are scripted), so the fix is to disable the resource check if Space Age is not enabled.

ClaudeMetz:
Well it's not really fixed for the next release, but the check is disabled entirely for now, and I'll look into custom integration with SE to properly address it.

PR #833, opened 2026-08-18
Fix surface compatibility

vapopescu:
Fix for #832


ClaudeMetz (commented):
Eh this is an incomplete fix, it's not really tied to modded planets, just surfaces where resources are scripted. It's why I was weary of this feature in the first place. I think the proper fix for this is probably a compatibility interface where SE tells FP which resources it generates manually. I'll look into that.

vapopescu:
Yes, that's right. But I think having all resources enabled in the mean time is better than having them disabled.

ClaudeMetz (commented):
Yeah I agree, I'll disable the check fully for now.

PR #835, opened 2026-08-20
Implement `SimpleItem:pack()`

vapopescu:
Motivation:
- Move backend logic from the handler to the backend
- Allows using `SimpleItem` instead of `FPItemPrototype` to describe Gaussian matrix free items (prerequisite for solver quality)


ClaudeMetz (commented):
Yeah this is much nicer, thanks.

Issue #854, opened 2026-08-27
Unaggregated recipes in Nullius


drakesiard:
### Problem Description

Nullius and Factory Planner have issues with fixed-recipe machines. When trying to fulfil the requirement for a seawater output, it shows two recipes rather than one, and each only has one machine available.

[factory-planner-nullius-bug.zip](https://github.com/user-attachments/files/31511201/factory-planner-nullius-bug.zip)

### Reproduction

1. With Nullius 2.0.11 and Factory Planner 2.0.50 (though the versions are not important), open Factory planner
2. Add seawater as a desired output
3. Click on the seawater output to add a recipe <img width="384" height="374" alt="Image" src="https://github.com/user-attachments/assets/5a5f0a57-f4f4-42b9-8988-07570d1be7a0" />
4. Note that there two identical "recipes" available, one for each machine 
5. Select the machine for the recipe <img width="288" height="244" alt="Image" src="https://github.com/user-attachments/assets/84595e4c-27f6-4537-ae12-9e49b7ecbd5c" />
6. Note that there is only one machine available for the recipe (instead of two)

drakesiard:
The recipe in question is nullius-seawater, e.g. https://github.com/GregorSamsanite/nullius/blob/29ad5b15d50b82b4e96c2f53229ec27c2b202cd8/nullius/prototypes/entity/plumbing.lua#L36 and https://github.com/GregorSamsanite/nullius/blob/29ad5b15d50b82b4e96c2f53229ec27c2b202cd8/nullius/prototypes/item/fluid.lua#L1917

ClaudeMetz:
Thanks for the great report, fixed for the next release.

PR #855, opened 2026-08-27
De-duplicate fixed recipes

vapopescu:
Fix for #854

Tested with Nullius: seawater, freshwater, air
Tested with Pyanodon: incubated petri dish, clay, proto tholins
Tested with AAI Industries (Space Exploration): processed fuel (single machine)
Tested with Ultracube: ionized annihilation stream (single machine)

### DIFF
--- a/modfiles/backend/handlers/generator.lua
+++ b/modfiles/backend/handlers/generator.lua
@@ -195,17 +195,13 @@ function generator.recipes.generate()
             if first_generator == nil or proto.name < first_generator then first_generator = proto.name end
         end
 
-        -- Recipes fixed to machines are duplicated with a special category
+        -- Recipes fixed to machines are given a special category
         if proto.crafting_categories and proto.energy_usage and proto.fixed_recipe then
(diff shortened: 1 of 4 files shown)


vapopescu:
Maybe also check if de-duplicating rocket silo launch recipes is possible/worth it.

ClaudeMetz (commented):
How did you test with nullius, seeing as it's 2.0 only currently?

vapopescu:
I used this fork https://github.com/sstrang/nullius

ClaudeMetz (commented):
Thanks for the fix, nicely solved, even if these kinds of fixes require stupid migrations. I have an idea for the silo to improve things, I'll do that separately though.

ClaudeMetz:
Not sure if it's what you meant, but I did some deduping of silo recipes along the same pattern as boilers etc.

vapopescu:
I would've used the old disambiguator as a suffix, but otherwise it's pretty much what I meant. Nice!

Issue #856, opened 2026-08-28
Simplex not solving for certain recipes


ClaudeMetz:
I ran into this in SA with the special rocket launch recipe:

<img width="1080" height="245" alt="Image" src="https://github.com/user-attachments/assets/523b6e41-4da9-4f0b-9242-0cf9c7d1c5a1" />

Simplex solved this one to 0. This product is a special entity one, but there could/will be others, might be because of that?

@vapopescu

ClaudeMetz:
Fixed for the next release.

PR #857, opened 2026-08-28
Fix simplex special item cost

vapopescu:
Fix for #856

### DIFF
--- a/modfiles/backend/calculation/simplex_engine.lua
+++ b/modfiles/backend/calculation/simplex_engine.lua
@@ -24,8 +24,8 @@ local simplex_engine = {}
 -- The objective function is maximized, so positive values indicate a score,
 -- and negative values indicate a cost
 local objective_vector = {
-    target_product = 1e9,
-    target_machine = 1e9,
+    target_product = 1e15,
+    target_machine = 1e15,
     limited_ingredient = 0,
     product = 0,
     ingredient = -0.001,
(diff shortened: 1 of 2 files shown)


ClaudeMetz (commented):
Does the change to the lib function make sense? It adds the heating energy one which should be the same as heat right?

vapopescu:
I guess, but why are there 2 types of heat? We will eventually want to solve for Aquilo heat requirements too, right?

ClaudeMetz:
heating is what buildings need to keep running on Aquilo.

vapopescu:
I see. But my question is whether when heating towers are implemented. Will you be able to produce heat for heat exchangers or heating for buildings. Or a recipe for either of them?

In my mind they are the same thing, that's why I'm a bit confused.

ClaudeMetz:
Oh I see, I guess that is a bit awkward. I wanted to split it out on recipes so it's clear what is the energy for thawing it and what is energy to run it. But it would be awkward when adding heat producers. I'll have to think about how to handle that.

vapopescu:
Ohh, so kinda like fuel then. Well it's something to think about.

We could treat them the same in the solver, then separate them again in the UI like we do with the fuel, if that's what you end up going with.

ClaudeMetz:
Yeah that sounds like the way to go, if we want to keep the separation. Maybe I don't care enough and we'll drop it, not sure yet. Maybe it could just be visual, so it's all treated as heat, but the icon would be normal heat if it's feeding the machine, or the heating icon if it's used for heating, and either one if it's both. We'll see, heat recipes have more complicated things to work out, like neighbor boni 😅

Issue #859, opened 2026-08-31
Save load crash


ClaudeMetz:
This error happens when loading this save

<img width="654" height="611" alt="Image" src="https://github.com/user-attachments/assets/302f0859-61f5-47db-bcd8-881bcb74296f" />

File 'PyAE V. 2 - 3nd Run.zip' in the drive

vapopescu:
The bug is caused by a TURD recipe replacement.

During migration, the factories are updated with the original recipes. But by the time the UI is refreshed, the recipe has been modified.

The culprit in this case is 'Moss MK1' TURD dropping 'Muddy sludge' (a fluid) from the recipe. Since 'Muddy sludge' is no longer an ingredient, the temperature data for it has been dropped.

ClaudeMetz:
Thanks for the analysis, spot on. I decided to do a 'proper' fix instead of yours, seen in 7091f7eb3bdc5cf4acf1763bc9709616ee37960e. Fixed for the next release.

PR #862, opened 2026-09-01
Fix temperature data missing

vapopescu:
Dirty fix for #859.

The proper fix would be to do the "on configuration change" solving step only after TURDs have been applied.


ClaudeMetz:
Proper fix in 7091f7eb3bdc5cf4acf1763bc9709616ee37960e, thanks.

PR #865, opened 2026-09-02
Fix solving for inactive floors

vapopescu:
Before the fix, disabling the relevant line would sometimes fail to disable the whole floor for the simplex solver.


ClaudeMetz:
Thanks, ran into this and wanted to make a report before seeing you already fixed it!

Issue #798, opened 2026-07-31
Support for reactor recipes


ClaudeMetz:
Reactors are heat producers that converts electrical or chemical energy to heat, much in the same (?) way a generator does. So the basic structure should be mostly given there. Needs a singular heat recipe in addition to the machines.

The main difference is the neighbour bonus. It's something that depends on the physical layout, and thus user configuration. How that is offered is kind of tricky, but I'm sure I'll find a way that works for most people. 

I also need to find a solution to the heat/heating split I introduced. Both are fed by heat, but they are different items, which makes it a bit messy. There's a couple approaches possible. One could combine them to feed into the solver, and split them out again afterwards. Alternatively, one could make them one item, and differentiate them visually with a different icon for display only. Or just make it one item and care about the difference entirely.

PR #867, opened 2026-09-02
Fix simplex heat/heating

vapopescu:
Now the solver treats them as separate items, just like the other solvers.

We need to decide how/if we merge them when implementing heat generation pseudo-recipes.


ClaudeMetz:
Thanks, I noted the issue in #798

PR #871, opened 2026-09-11
Fix merge error

vapopescu:
Minor bug squeezed in while merging #865


ClaudeMetz:
Ugh I really struggled with that merge, brain too tired should have stopped earlier lol thanks.

PR #873, opened 2026-09-11
Matrix solvers util refactor

vapopescu:
For both matrix solvers:
- Re-use the same set operations (union, intersect, subtract) between the matrix solvers

For Gaussian matrix solver:
- Change the solver to use the solver item keys instead of its own keys
- Reduce redundant set -> list -> set conversions
- Remove machine count workaround for recipes with 0 energy (it's handled in `solver.set_line_result()`)


ClaudeMetz:
Nice, I see you've been stacking up the changes while I'm being slow to merge them :p 

What about the `print_` functions in the Gaussian solver btw? I don't really love 'debug code' in the codebase, so if you don't use them we can consider dropping them.

vapopescu:
Yeah. I've been hoarding for a while :D. This is the last one directly related to per-floor solving though, so I'll keep the quality related ones closed for now.

We can get rid of the prints, sure. With this change, the item keys are very readable, so the debugger variables should be pretty readable too. The recipe keys still use integer ids, but plugging the id into `storage.prototypes.recipes` should be easy enough.

ClaudeMetz:
Alright sounds good, thanks as always.

ClaudeMetz:
When you see +146 -319 on a refactor PR, you know it's a good one, nicely done.

PR #882, opened 2026-09-12
Fix entity ingredients appearing as items

vapopescu:


ClaudeMetz:
I kind of dis this change on purpose, but I guess it's cleaner to give it a different background yeah.

vapopescu:
Sorry, I thought it was accidental, given that the compact dialog still used the old style. But you decide.

ClaudeMetz:
Yeah then the compact dialog was the mistake lol but I think you're right, they should be a different color since they aren't really the same thing.

PR #884, opened 2026-09-13
Gaussian matrix data refactor

vapopescu:
@scottmsul FYI

- Merge matrix metadata, factory metadata and matrix lines metadata into a single structure
- Cache the necessary data for the production box in the backend, and remove the dependency on matrix engine calls and metadata
- Move the linear dependency check from `run_matrix_solver()` to `get_linear_dependence_data()`
- Use semicolon as key separator
- Use recipe names and line IDs for their respective keys


ClaudeMetz (commented):
Moving the complicated checks out of solver.lua for the gaussian solver is huge, thanks a lot. The rest is not too bad either I'd say, as far as I can tell at least :)

Issue #886, opened 2026-09-14
Another strange 'no product requested' problem for the simplex solver


breadje:
### Problem Description

I got another strange 'no relevant product requested' bug with the simplex resolver. I need 184 acetylene (@therenas per second! ;) ). But when I click it to ad the production (there is only one recipe) it gives the above error. The factory does not use the slacked lime by product as far as I can see and only one place needs the acetylene. 

<img width="1553" height="1025" alt="Image" src="https://github.com/user-attachments/assets/d3664a9c-2f1a-46b2-bac2-2b4f32d3cc1d" />
<img width="1549" height="1019" alt="Image" src="https://github.com/user-attachments/assets/15721f1a-e7e9-4941-b55c-e57c6fc64bb8" />

breadje:
Factory export:

[...]

PR #887, opened 2026-09-14
Normalize simplex tableau before solving

vapopescu:
Fix for #886.

Now the simplex solver should be more resistant to tolerances.


ClaudeMetz:
Thanks.

PR #888, opened 2026-09-14
Common solver input data

vapopescu:
- Change all solvers to use line aggregates assuming one machine as input data
- Rename matrix solver to Gaussian solver

This one is also non-breaking, but given the size, I would suppose you want to merge this on a separate feature branch.


ClaudeMetz:
Jeez, big one! By separate branch you mean not release it and accumulate changes until per-floor solvers are done?

If so I'd prefer to just merge stuff to master as much as possible, and only do the actual breaking changes in a branch. It's a bit risky because there could be bugs of course, but 2.1 still being experimental means not too many people are running it, and we can do hotfix releases very easily.

vapopescu:
Cool. Then you can review and merge at your pace.

ClaudeMetz:
I haven't fully reviewed it yet, but this change is amazing. This kind of centralize as much as possible-approach is something I've wanted to do since forever but I didn't dare touch the gaussian solver so I never did.

ClaudeMetz:
One thing I ran into and I'm not sure about is that simplex likes to scale recipes higher if their fuel can consume a byproduct, like this contrived example:

image

I don't know if it makes sense to scale a recipe just to use up the fuel. I guess there's situations where this makes sense though. Would this be adjustable to not use all the coal but just the iron if we had simplex item priority adjustement?

vapopescu:
It's a bug. Simplex cannot export chosen products as byproducts. You can open an issue

ClaudeMetz:
Alright will do. Another thing with the sequential solver: It can scale consuming recipes by fuel instead of main product. I feel like it should always scale to main product as a priority. Example is of where it goes wrong. It should consume all iron ore instead of limiting itself to the fuel.

image

ClaudeMetz:
> It's a bug. Simplex cannot export chosen products as byproducts. You can open an issue

Actually, I don't understand. I'm not saying it's doing anything wrong necessarily. What's the bug?

vapopescu:
>  It can scale consuming recipes by fuel instead of main product.

Oh right. Before the change, things like fuel, electric power and heat did not exist in the line aggregate when making the scaling choice. Adding an exception just for fuel would be sufficient?

> I'm not saying it's doing anything wrong necessarily. What's the bug?

It's half bug, half limitation. After I thought about it more, it should be fixed when defining products by machine limits.

As it is currently, simplex adds two types variables for each item. One is item import (ingredients and some intermediates), and the other is item export (products, byproducts, and intermediates).

For desired products, the output variable for that item is given a different objective coefficient, so instead of getting punished for exporting that item, it is rewarded instead. But at the same time, a condition is put in place so that the export amount cannot exceed the amount specified by the user (in your case, 60 coal and 60 iron ore).

The bug/limitation is that solutions where a desired product is exported more then what the user specified are considered invalid. If you didn't have the iron plate line to consume them, it would not give a solution at all, as no valid solutions exist (fixed 3 miners that output 60 coal/s or less). This is why products defined by machine limits would fix this, since you would have only one of these conditions in place.

vapopescu:
> Would this be adjustable to not use all the coal but just the iron if we had simplex item priority adjustement?

If the coal is not a desired product like in your example, then yes.

vapopescu:
image
image

This looks ok, right?

ClaudeMetz:
Yeah that looks right, using all the byproduct is the desired solution I think. It's unavoidable that the top level product isn't fully satisfied sometimes, there's other scenarios where it happens. That's where you should use simplex, but yeah. So that plus the other PR fixes the two issues right?

vapopescu:
The simplex issue is two-fold, and the other side will be fixed by user import/export priority on intermediates. But yeah, unless you find something else this one is ready.

ClaudeMetz:
I wanted to get this into the release, but I had to rush it for a bug fix and I wasn't quite done reviewing. I'll probably get it merged tomorrow.

ClaudeMetz:
Well I haven't looked into every nook and cranny but let's merge this, great stuff as always.

PR #892, opened 2026-09-16
Make simplex be able to overproduce products

vapopescu:
Related to the discussion in #888

Improve on simplex limitation which previously did not consider over-producing a product as a valid solution.


ClaudeMetz:
Thanks for the quick work.

PR #894, opened 2026-09-17
Fix crash when exporting disabled subfloor

vapopescu:
A disabled subfloor has no products. So take the zeroed products of the first line instead.


ClaudeMetz:
Thanks!

PR #895, opened 2026-09-17
Per floor solving

vapopescu:
Will change target branch when it's created.

Stuff that breaks:
- Machine limits on floor 2+ for sequential solver
- Existing Gaussian solver solutions for multi-floor factories

New stuff:
- All solvers are solved per-floor in a common framework
- Sequential and Gaussian `LineResults` now only pass the solved machine amount. The resulting line aggregates are calculated on a common path
- Machine limits for Gaussian solver for the 1st floor and adjacent lines (top line of floor 2)
- Auxiliary solver data is also stored per-floor

I re-ordered functions in `solver.lua` and `gaussian_engine.lua` in order to make them local. As a result, the changes are quite hard to read. Sorry for that.


ClaudeMetz:
I'm working on the machine limit move, which it seems is quite relevant to this and integrates well from what I can see. I'll finish that up first, then we can incorporate it here and go from there. You'll get to review that before it gets merged of course.

vapopescu:
I implemented my idea of floor status and solver errors. You can check and change as needed.

ClaudeMetz:
Nice, I'll do the settings and stuff and then review, might take a few days.

ClaudeMetz:
So how does the migration of the gaussian solver work in practice? I guess if there's multiple floors there'll be too many free items in most cases and people will need to unselect some to get things to work again?

vapopescu:
This is pretty much correct. The solver also filters free items that only appear on that floor, so the number is also reduced that way.

Since this is a one-time thing, I did not want to do anything more complicated than this. But if you have other ideas, let me know.

ClaudeMetz:
No that sounds about right, it's unavoidable that this breaks some setups, and I'll mention that in the changelog.

So technically this doesn't implement per floor solver choice, so we could merge it without that, but it's just a GUI thing to do that right, so we should do it all at once?

The solver architecture if beautiful btw, distilling the solvers as much as possible to what makes them unique and consolidating everything else in one file. It's what I always had in mind but never had the skills to achieve :)

vapopescu:
Thanks! I appreciate the compliments

> So technically this doesn't implement per floor solver choice, so we could merge it without that, but it's just a GUI thing to do that right, so we should do it all at once?

Your choice. I'll do a small change to prepare the per-floor setting. If everything works well, you only need to change that and it will work.

ClaudeMetz:
Btw this crashes when starting a new game, I could have a look at a fix but you probably know the right fix. Go ahead with the preparation too then, we'll do the changes all at once.

vapopescu:
Done with the change and the crashes.

ClaudeMetz:
I moved the warnings to the solver bar, as the production bar is a bit too crowded on smaller displays potentially.

ClaudeMetz:
Please have a look at whether the pref makes sense to you and whether the changelog is correct and complete. (I still have reviewing to do but wanted to make my changes now)

modfiles/changelog.txt
```diff
@@ -2,6 +2,9 @@
 Version: 0.00.00
 Date: 00. 00. 0000
   Features:
+    - Rewrote the Gaussian solver to bring it in line with the other solvers. It now treats subfloors in isolation instead of internally combining everything onto a single one. This will break some setups and/or require reconfiguration. The solver now also supports machine limits. (#842) (Thanks vapopescu!)
```
vapopescu: I think this is a change, not a feature. Added machine limits are a feature.

modfiles/changelog.txt
```diff
@@ -2,6 +2,9 @@
...
+    - Added a setting to allow you to use distinct solvers for different floors when desired. As a consequence, solver configuration is now always done per floor, not globally per factory (#768) (Thanks vapopescu!)
+    - Reworked how errors related to the solvers are presented to improve clarity about what is going wrong (Thanks vapopescu!)
```
vapopescu: Also a change

vapopescu:
The Gaussian solver status looks a bit weird being split into 2 labels. I will adjust it. Adding tooltips to explain the errors may also not be a bad idea.

Other than these, it looks good.

Issue #343, opened 2024-10-19
Show which planets an incompatible recipe can be crafted on


ClaudeMetz:
Currently the UI just shows when a recipe or machine is not compatible with the current planet. That's fine, but it would be neat to show the planet it does work on. This has a few technical challenges.

The cleanest solution would be to precompute a list of compatible planets for every combination of recipe and machine. That would be a crazy large number of combinations, were it not for the fact that most recipe/machines don't have surface conditions. Still, this might prove to be too much.

Alternatively, you could precompute compatible planets for each recipe and machine separately, and combine their compatibility on demand (which is then cached). This is less clean, but more performant.

vapopescu:
I would lean on the second option. Gather the surface compatibility data on the first generator pass, then compute the location compatibility sets separately for recipes and machines in their respective second passes.

ClaudeMetz:
Interesting. What I meant is compute the intersection when the user hovers the tooltip, not during generator at all. But I haven't done the math on the combinatorics here.

vapopescu:
This is also what I meant. Compute and store separately in recipe and machine prototypes, then intersect later on-demand.

PR #897, opened 2026-09-17
Simplify location compatibility calculation

vapopescu:
Improvement towards #343.

- Compatible/incompatible locations are stored in the recipe and machine prototypes.
- Lines only cache the current location.


vapopescu:
I also moved resource restrictions to the generator second pass with this change. This time, it only applies the restrictions if the resource can be found on at least one planet. This way, it should not disable resources that are only added to the map with scripts (i.e. Space Exploration resources).

If you still want to disable resource restrictions anyway, you can do so from `generator.lua:670`.

PR #902, opened 2026-09-20
Implement defining products via machine amount

ClaudeMetz:
This replaces the limits on machines with a limit on factory products. The limit itself works in the same way, it's just being applied from the product down to the first matching recipe.

Removing the limits on machines is technically a regression for some use cases, but we could always bring it back if we think we really need to.

@vapopescu Please have a look, just to verify this'll work with everything. I didn't touch the matrix solvers yet. I realize this intersects with the per-floor solvers PR, so I wonder what the less painful way to go about this would be: Complete the implementation on this branch for the other solvers, merge into master, merge into the other branch and go from there, or just target the other branch onto this and work everything out in one go. Your choice, I don't mind it either way.


vapopescu:
You can merge, and I'll re-implement the thing in my PR.

ClaudeMetz:
I can't quite merge it this way since it'll be broken for the matrix solvers, and I might do a release before the other branch is merged 😅 I could just fix it so it's ignored in the other solvers for now though.

